PR #23, opened 2024-07-29
PR: Analytics runbooks are scheduled at incorrect times

jesusfer:
Runbooks schedules should be created at specific times but all are created at the same time.

Fix:
- Use the proper times when creating/updating the runbook schedules.


jesusfer (commented):
Updated schedules and added tests.

PR #52, opened 2025-05-27
Duplicate URL fix for importing AppInsights data

sambetts:
## Duplicate URL Clean-up Script
Due to max-length varchars not having uniqueness constraints in the database, duplicate URLs were being imported into the database under some specific conditions. [SQL Server doesn't allow indexes over 900 bytes](https://learn.microsoft.com/en-us/sql/sql-server/maximum-capacity-specifications-for-sql-server), and shrinking the full_url field on "urls" table isn't an option. 

## Type of Change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
This PR is mainly just a script that can be run if there are duplicate URLs. 
For each URL lookup that have +1 same URL string in full_url, we migrate all lookups to the 1st found result, then delete the old lookup.

Script is: at "src\AnalyticsEngine\WebJob.AppInsightsImporter.Engine\Resources\Delete Duplicate URLs.sql"
Tests: src\AnalyticsEngine\Tests.UnitTests\DuplicateUrlTests.cs - ensures the right data is deleted by the T-SQL.

### DIFF
--- a/src/AnalyticsEngine/WebJob.AppInsightsImporter.Engine/Resources/Delete Duplicate URLs.sql
+++ b/src/AnalyticsEngine/WebJob.AppInsightsImporter.Engine/Resources/Delete Duplicate URLs.sql
@@ -28,7 +28,7 @@
 							GROUP BY full_url
 							HAVING COUNT(*)>1
 						) dt ON y.full_url=dt.full_url
-						order by dt.CountOf
+						order by dt.full_url
 					
 		open @allUrlsCursor
 


src/AnalyticsEngine/WebJob.AppInsightsImporter.Engine/WebJob.AppInsightsImporter.Engine.csproj
jesusfer, line 1: Why is this one Embedded Resource instead of Content?

PR #101, opened 2026-05-28
New v3 for Analytics Report

jesusfer:
## Title

Updated Analytics Report.

## Type of Change

- [ ] New feature
- [ ] Bug fix
- [x] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description

This PR aims to update the Analytics report and make it lighter in the Power BI service by:
* Reducing the number of tables loaded in the data model. Remove the unpivoted activities table from the model.
* Use query folding in Power BI to join tables in SQL rather than in Power Query.
* Add incremental refresh to the metrics tables.
* Use measures in the matrix and not an unpivoted table to help the engine work faster.

## Breaking changes

Once this PR is merged:
* Profiling stored procedures will stop updating the profiling.ActivitiesWeekly table so previous versions of the report will be degraded.
* Customized reports should be updated to use the new data model provided measures and tables. Many will have been removed.

## Related issue(s)

- N/A

## Check list

- [ ] Related issue / work item is attached
- [ ] Changes are tested
- [ ] Tests are written (if applicable)
- [ ] Documentation is updated (if applicable)

## Pending changes
- [x] Update the report data model
  - [x] 1. Activity metrics
  - [x] 2. Usage metrics
  - [x] 3. Example KPIs
- [ ] Update SQL stored procedures


sambetts:
Triage note from #275.

Thanks for this report-model v3 work. Current state from GitHub:

- Draft PR, opened 2026-05-28 and last updated 2026-07-21.
- Targets `dev` and is currently mergeable, but blocked because it is still draft.
- Scope is additive: new `Analytics_DataModel_v3_base.pbit`, new `dax_helpers.ipynb`, and README updates.
- The PR body says the intended report model removes the unpivoted activities table, relies more on SQL/query folding, adds incremental refresh, and moves matrix logic into measures.
- It also says SQL stored-procedure updates are still pending and that older/custom reports may be affected because `profiling.ActivitiesWeekly` would stop being updated.

This looks potentially salvageable, but it should not land as-is while the SQL/profiling side is still marked pending. To revive it, I think it needs either:

1. a completed end-to-end PR with the required profiling SQL/report-model changes, refresh validation, and upgrade documentation; or
2. a narrower split where the PBIT/helpers/docs are kept as a non-breaking preview and the profiling change is handled separately.

Could you let us know whether you want to bring this up to date and mark it ready for review, or whether you would prefer to close it and reopen a smaller replacement later?

sambetts:
@jesusfer — cross-linking this to **#358**, which you've just been asked to review, because it changes numbers your v3 model consumes.

#358 fixes a date-boundary bug in `profiling.usp_UpsertCopilot` (the proc you originally wrote in c022069). The weekly window was effectively losing its last day, so once it lands **weekly Copilot figures compiled by that proc go up by roughly 1/7 (~14%)**. Two consequences for this PR:

- Any baseline, expected value or screenshot in the v3 model derived from current weekly Copilot data will shift.
- Historical weeks stay under-counted until they're recompiled, so a tenant will see a step change at the release boundary rather than a uniform correction. Whether recompilation is in scope is one of the open questions on #358 — your view would help settle it.

No action needed on this PR itself. For the record, since "339 commits behind `dev`" looks alarming out of context: it isn't a problem here — this branch only touches `reports/Usage Analytics/`, so it still merges cleanly against current `dev`. It's marked draft, so I've left it alone.

PR #128, opened 2026-06-16
Fix oversized Teams deep-link (xsdata) URLs blocking urls.full_url migration

### LINKED ISSUE #127: Oversized Teams deep-link (xsdata) URLs block urls.full_url migration and fail imports
## Problem

SharePoint links opened from Microsoft Teams carry a large, **transient** `xsdata=` query token
(plus `sdata`, `ovuser`, `clickparams`, …). These URLs routinely exceed the new `dbo.urls.full_url`
width (`nvarchar(850)`, see #122 / migration `ShrinkUrlsFullUrlColumn`), which:

1. **Blocks the `ShrinkUrlsFullUrlColumn` upgrade** — the migration aborts when any `full_url` is
   longer than 850 characters. Real customer samples are **1213–1515** characters.
2. **Fails imports going forward** — the activity / click / Copilot importers stage URLs into
   `nvarchar(850)` staging columns; an over-length value throws `BatchSaveException` (`InsertBatch`
   does not truncate).

The `xsdata` token alone is typically 600–1100 characters of opaque base64 with **no analytic
value**. Removing just `xsdata` brings the observed samples down to **601–935** characters.

## Fix

- **Upgrade guide** (`ShrinkUrlsFullUrlColumn-UpgradeGuide.md`): add (a) queries to count total vs
  oversized URLs (and how many carry `xsdata`), and (b) a clean-up script that strips `xsdata` from
  oversized rows. The script only applies a cleaned value when it **fits** (≤ 850) **and does not
  already exist** (so it never creates a duplicate `full_url`), collapses rows that clean to the
  same value, and is wrapped in a transaction that **defaults to `ROLLBACK`** (preview first, then
  flip to `COMMIT` to apply).
- **Importers**: strip `xsdata` at insert time when a URL exceeds the column width, and
  hard-truncate to 850 as a last resort, via new `StringUtils.EnsureUrlWithinLength`. Applied to
  audit `ObjectId`, click `HRef`, and Copilot `Url` staging.

## Notes

- All text stays Unicode-safe (`nvarchar`, never `varchar`).
- Validated: cleaning logic 15/15 (5 real customer URLs + 10 edge cases) cross-checked Python ↔
  T-SQL; the clean-up script verified end-to-end on localdb (rollback no-op, commit cleans the right
  rows with zero duplicates, skips collisions, leaves still-oversized rows); new unit tests pass.

sambetts:
Fixes #127.

## What

Oversized SharePoint **Teams deep-link** URLs (carrying a large transient `xsdata=` token) were
blocking the `ShrinkUrlsFullUrlColumn` migration (`dbo.urls.full_url` → `nvarchar(850)`) and would
fail imports into the new `nvarchar(850)` staging columns. Removing just `xsdata` brings observed
real samples (1213–1515 chars) back under 850 (601–935).

## Changes

**Docs — `ShrinkUrlsFullUrlColumn-UpgradeGuide.md`**
- Count queries: `total_urls`, `oversize_urls`, `oversize_with_xsdata`, `oversize_other`.
- A clean-up script that strips `xsdata` from oversized rows. It only applies a cleaned value when
  it **fits** (≤ 850) **and does not already exist** (never creates a duplicate `full_url`), and
  collapses rows that clean to the same value. Wrapped in a transaction that **defaults to
  `ROLLBACK`** — runs as a preview, admin flips to `COMMIT` after verifying.
- A "what's left" section for rows still > 850 after stripping or that collide.

**Importers**
- New `StringUtils.RemoveXsDataParam` (boundary-aware, case-insensitive, preserves the rest of the
  URL incl. `#fragment`) and `StringUtils.EnsureUrlWithinLength(url, max)` — strips `xsdata` only
  when over `Url.FullUrlMaxLength` (new `850` const), then hard-truncates to 850 as a last resort so
  staging never overflows.
- Applied at audit `ObjectId`, click `HRef`, and Copilot `Url` insert points. (`HitTempEntity`
  already drops the whole query string.)

## Validation

- Cleaning logic 15/15 (5 real customer URLs + 10 edge cases) cross-checked Python ↔ T-SQL.
- Clean-up script verified end-to-end on localdb: rollback is a no-op; commit cleans the right rows
  with **0 duplicates**, skips collisions, leaves still-oversized rows, ignores short / non-xsdata
  URLs.
- Solution builds; 2 new unit tests pass.

## Notes

- All text stays Unicode-safe (`nvarchar`, never `varchar`).
- Hard-truncation can rarely make two distinct degenerate URLs share an 850-char prefix and merge
  into one `urls` row — acceptable for junk Teams tracking URLs, flagged for awareness.

### DIFF
--- a/src/AnalyticsEngine/Common/DataUtils/StringUtils.cs
+++ b/src/AnalyticsEngine/Common/DataUtils/StringUtils.cs
@@ -207,13 +207,21 @@ namespace DataUtils
         }
 
         /// <summary>
-        /// Ensures <paramref name="url"/> fits the <c>dbo.urls.full_url</c> column. URLs already
-        /// within <paramref name="maxLength"/> are returned unchanged (so we avoid needless churn).
-        /// Otherwise the volatile <c>xsdata</c> parameter is removed (see <see cref="RemoveXsDataParam"/>);
-        /// if the result is STILL too long (e.g. a huge <c>FilterValues</c> list, or a URL with no
-        /// <c>xsdata</c> at all) it is hard-truncated to <paramref name="maxLength"/> as a last resort,
-        /// so the importer never fails staging an over-length value into the nvarchar(850) staging
-        /// column. Pass <c>Common.Entities.Url.FullUrlMaxLength</c> as <paramref name="maxLength"/>.
+        /// Ensures <paramref name="url"/> fits the <c>dbo.urls.full_url</c> column, mirroring the
+        /// staged clean-up in ShrinkUrlsFullUrlColumn-UpgradeGuide.md so the importer can never fail
+        /// staging an over-length value into the nvarchar(850) staging columns. URLs already within
+        /// <paramref name="maxLength"/> are returned unchanged (no needless churn). Otherwise, in order:
+        /// <list type="number">
+        /// <item>the volatile <c>xsdata</c> token is removed (see <see cref="RemoveXsDataParam"/>) -
+        /// on its own this brings the large majority of oversized SharePoint URLs back under the limit;</item>
+        /// <item>if it is STILL too long (e.g. a giant <c>SafelinksUrl</c>/<c>FilterValues</c> list or
+        /// comma-merged Teams parameters), the whole query string and <c>#fragment</c> are dropped,
+        /// keeping just the page path - the only part with analytic value (this matches the "reduce to
+        /// path" step of the upgrade guide);</item>
+        /// <item>only in the pathological case where even the bare path exceeds <paramref name="maxLength"/>
+        /// is it hard-truncated as an absolute last resort.</item>
+        /// </list>
+        /// Pass <c>Common.Entities.Url.FullUrlMaxLength</c> as <paramref name="maxLength"/>.
         /// </summary>
         public static string EnsureUrlWithinLength(string url, int maxLength)
         {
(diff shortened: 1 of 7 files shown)


sambetts:
Expanded the upgrade guide's clean-up section based on a real upgrade where ~99% of oversized URLs were fixable by the `xsdata` strip and only a handful were genuinely stuck:

- **Section 3 — classify what's left:** a read-only diagnostic that buckets every remaining oversized row as *would-clean-with-COMMIT* / *still-too-long-after-xsdata* / *cleaned-value-already-exists* / *no-xsdata*, so the operator knows exactly what (if anything) still needs hand-resolution after the preview is committed.
- **Section 4 — reduce genuinely-stuck rows to their path:** a `ROLLBACK`-default de-duplicating merge for the rows that even `xsdata` removal can't bring under 850 (big `SafelinksUrl`/`FilterValues`/comma-merged Teams junk). It reduces each to its page path, repoints **every** foreign key that references `urls` (discovered from the catalog, so no referencing table is missed) onto a single canonical row, then deletes the duplicates — so it never creates a duplicate `full_url` and never loses referenced facts. It deliberately leaves rows the `xsdata` strip can already fix untouched.

All five SQL blocks in the guide were extracted from the committed markdown and run end-to-end on SQL Server localdb (FK repoint, dedup, rollback-as-no-op, commit-fixes-everything). All examples are anonymised.

PR #157, opened 2026-06-24
Modernise the admin web app: Vite/React 19 + Fluent UI, User Data Lookup, optional Redis, merged Home + Install Log

sambetts:
## Overview

Modernises and substantially expands the analytics **Web** admin app, makes the web app run **without Redis**, and gives it an **Office 365 look & feel**. Builds on top of the original Teams-permission tool.

### 1. Rename + modernise (`teams-permission-grant` → `admin-app`)
- Migrated the deprecated **Create React App** project to **Vite + React 19 + TypeScript**; upgraded `@azure/msal-browser` to v5.
- Adopted **Fluent UI React v9** (`@fluentui/react-components`) across the whole SPA — Office 365-style header, `TabList` nav, Fluent controls throughout. Bootstrap removed.
- Multi-page SPA via `HashRouter`, served by a single `[Authorize]`'d MVC action.

### 2. User Data Lookup (new)
- Admins enter a **UPN** and see all of that user's data held in SQL: profile, per-category record counts, and drill-down to recent rows.
- Counts are **broken down by workload** — including **Copilot interactions** and the other audit sub-types (SharePoint, Exchange, Entra, Power Apps/Automate/BI, Copilot Studio), plus sent emails, web hits, Teams membership/owners/reactions/calls/feedback, page likes/comments, app installs, usage reports and Power Platform shares.
- Each row exposes the **SQL behind its count** (a popover with a copy-to-clipboard button) so an admin can verify it in SSMS.
- Shows which **import workloads** are enabled, with an "import off" badge, so a legitimate 0 count is explained rather than looking like a fault.
- Scale-safe for ~200k-user tenants: direct UPN equality compare (no `ToLower()` scan), indexed FK counts, bounded `Take` on every detail query.

### 3. Redis is now optional for the Web app
- Previously the site **crashed at startup** with `ArgumentNullException` when no `Redis` connection string was configured.
- `CacheConnectionManager.TryGetConnectionManager(...)` returns `null` when Redis isn't configured; all web consumers degrade gracefully (the importer web-jobs still require Redis).

### 4. Fix: admin app "No server-side credentials" without Redis
- The Graph token was only ever stored in Redis, so with no Redis the SPA had no token.
- Server-cookie approach: during the AAD sign-in redirect the server stores the OAuth **refresh token** in the encrypted, httpOnly **auth cookie**; `SiteTokenAPI` mints a fresh Graph **access token** from it. Works with no Redis. (Teams deep-analytics tokens still live in Redis, since the separate importer process needs them.)

### 5. Home page merged into the SPA + new Install Log
- The old server-rendered system-status home page is now a React page (`api/SystemStatus`), the default route. `HomeController.Index` serves the SPA at `/`.
- The home overview now lists more of the main tables (Users, hits, audit events, Copilot, sent emails, Teams, calls, sites, URLs), cached briefly so the full-table counts don't run on every load.
- New **Install Log** page (`api/InstallLog`) showing the full `sys_configs` history — when each configuration was applied, by whom, install messages, and the config JSON (secrets are encrypted `*Hash` fields, not plaintext).

### 6. Tests
- Added unit tests for the user-lookup COUNT-SQL builder (direct-FK, web-hits and audit-event-join forms + UPN quote-escaping). The pure logic lives in `Common.DataUtils` so the test project doesn't need to reference the Web project. 6/6 pass.

## Database changes
**None.**

## Validation
- `admin-app` `npm run build` (type-check + Vite) — 0 errors.
- `Web` Release MSBuild (incl. the automatic `npm install` + `npm run build`) — 0 errors.
- `Tests.UnitTests` build + the new `UserDataLookupTests` — 6/6 green.
- **Runtime EF smoke test** against a freshly-migrated LocalDB: every new EF query in the PR (all User Data Lookup category counts + detail drill-downs, the audit-navigation joins `copilot_chats`/`event_meta_*` → `audit_events.user_id`, the multi-hop web-hits nav, the 9-`Include` profile query, and the Home/Install-Log queries) translates to valid SQL and executes against the real schema. A seeded `audit_event` + `copilot_chat` returns the correct Copilot count. See the smoke-test results comment.

## Not yet done (reviewer action)
The **interactive OIDC sign-in in a browser** (the `[Authorize]` endpoints + AAD redirect that populates the auth cookie) still needs one manual run-through on a real tenant before merge. The server-side token-mint and refresh-token claim round-trip were verified by static review; the EF audit-navigation SQL translation has now been smoke-tested (see Validation).

## Reviewer notes
- The auth cookie now carries the OAuth refresh token (claim `urn:aa:graph_refresh_token`) in the encrypted, httpOnly cookie — this is what removes the Redis dependency for the SPA's own Graph calls.
- The bundle is ~744 kB main + lazy-loaded route chunks (Fluent dominates) — acceptable for an internal admin tool.


sambetts:
## Smoke test results

Smoke tested locally (VS2022/18 MSBuild + LocalDB). Summary: **green, no bugs found.**

### Build
- Full **Release** MSBuild of the whole solution — Web (incl. the automatic `npm install` + Vite SPA build), both WebJobs, the installer, and `Tests.UnitTests` — **0 errors**.

### Tests
- Existing `UserDataLookupTests` (the COUNT-SQL builder) — **6/6 pass**.

### Runtime EF smoke test (the flagged risk)
Ran **every new EF query in the PR** against a freshly-migrated LocalDB (`UnitTestingAnalytics`) to validate SQL translation + execution against the real schema:

- **User Data Lookup** — all 31 category `COUNT` queries, every per-category detail drill-down (including the generic audit-child and usage-report projections), and the 9-`Include` profile query. All translate and execute.
- **The audit-navigation joins specifically** — `copilot_chats` / `event_meta_*` → `audit_events.user_id`, the multi-hop web-hits nav (`hits → session → user`), and `OrderByDescending(x => x.AuditEvent.TimeStamp)`. All good.
- **Correctness check** — seeded a user + `audit_event` + `copilot_chat` and confirmed the Copilot audit-nav lookup returns the right count (**1**), i.e. the join is right, not just translatable.
- **Home / SystemStatus** — the full-table counts including `Teams.Where(t => t.HasRefreshToken)` (confirmed a real mapped `has_refresh_token` column, not a computed property), and the **Install Log** `sys_configs` query.

### Redis-optional + auth-cookie token flow (static review)
Walked the flow end-to-end: `Startup.Auth` writes the `urn:aa:graph_refresh_token` claim into the encrypted cookie → `BaseAPIController.GetUserTokenFromClaims()` reads it → `SiteTokenAPI` mints a fresh Graph access token from the refresh token → Redis is a null-safe fallback throughout (`CacheConnectionManager.TryGetConnectionManager` returns `null` instead of throwing; `TeamsAuthAPI` returns a clear 503 when Redis is absent). Logically coherent and consistent.

### Not covered (needs a human + real tenant)
- The **interactive OIDC sign-in in a browser** (the `[Authorize]` endpoints + AAD redirect that populates the cookie) can't be automated here. The server-side token-mint and claim round-trip were verified by static review only — worth one manual sign-in before merge.

### Note
The runtime EF checks ran via a temporary test harness mirroring the controller's queries against LocalDB; I removed it afterwards, so the branch is unchanged. If a permanent regression guard is wanted, the clean way is to extract the count/detail `switch` into a small `Common` helper so a DB-backed test can call the real code without pulling a Web reference into the test project.

PR #212, opened 2026-07-14
Stable Release: large-tenant import & DB performance, opt-in Power Platform import, in-app Reports tab, private-deployment checkpoint fix, data-integrity + security fixes

sambetts:
## Stable Release

Promotes the current `dev` line to `main`. On merge, CI builds and signs the artifacts and creates a **draft "Stable build" release** with auto-generated notes; it stays a **draft until it's published**, so this is effectively a release candidate that can be tested first.

> ⚠️ **This release contains database schema changes (five migrations) and an installer config-schema bump (1.9.0).** Read **"Database upgrade steps"** (including the per-migration timing table) and **"Installer / config changes"** below before upgrading a production tenant. On a large, long-running tenant, budget roughly **1–3 hours** for the database upgrade and run it in a maintenance window with the importer stopped.

### New: opt-in Power Platform & Copilot resource-resolution import (#219, #220)
- The audit importer can now optionally import **Power Platform** activity and resolve **Copilot accessed-resource** metadata, controlled by **new per-import toggles** in the installer. Both are **off by default**, so existing tenants behave exactly as before until you turn them on.
- Because this adds a saved setting, the **installer config schema is now 1.9.0** — re-open and re-save your configuration in the installer (or re-run it) so the new option is written. Older config files still load.

### New in-app Reports tab (#201)
- **A new "Reports" tab in the admin web app** — a lightweight, in-app alternative to the Power BI dashboards that needs **no extra deployment**, giving operators a quick view of how Microsoft 365 usage is trending inside the admin site.
- **Each area only appears when its import is enabled** (Copilot, Microsoft 365 usage, SharePoint & OneDrive, Website traffic, Teams calls, Emails), so you never see an empty section. Every series is grouped by week.
- The **reporting window is configurable** (last 1 / 3 / 6 months, default 3), and each chart has a **"SQL" popover** so an admin can see and re-run the exact query behind it.

### Audit import on private / locked-down deployments (fix, #213)
- **The audit-import "blob checkpoint" now works on private deployments.** The checkpoint — which lets each import cycle skip content it has already processed — uses Azure **Table** storage, but the installer previously created a private endpoint for the storage **blob** service only. On deployments where the storage account blocks public network access, the table endpoint was unreachable, so the checkpoint silently fell back to a non-durable in-memory copy that was lost on every restart. That made the importer re-download the entire multi-day activity window each time — on large tenants, enough to stop an import cycle ever completing.
- The installer now provisions the matching **table private endpoint + private DNS zone**, the "Test Configuration" check validates it, and the networking wizard has new paired **Storage / Storage table** endpoint-name fields. Existing private deployments pick this up on an installer re-run (or can add the table private endpoint by hand).
- **Better diagnostics:** a degraded (non-durable) checkpoint now shows as an amber **BlobCheckpoint** card on the in-app Health page, the log records the actual reason durable storage couldn't be reached, a successful durable init is logged too, and unexpected errors are captured as proper Application Insights exception telemetry.

### Faster import & database on large tenants
- **Fixed the top database bottlenecks (#221).** User and site look-ups during import are now written so SQL Server can use its indexes (they previously forced full-table scans on every row), a **missing index on `sites(site_id)`** is added, and a **weekly whole-database index/statistics maintenance** job keeps things healthy. On a large tenant this is the single biggest CPU reduction in the release. *(Database schema change — see below.)*
- **Much faster audit-log import (#207).** The Office 365 audit importer now (a) adds the missing index on the audit-events time-stamp, (b) keeps a checkpoint of the content it has already downloaded so each cycle no longer re-downloads and re-parses the entire multi-day lookback window, and (c) saves to the database with more concurrency. *(Database schema change — see below.)*
- **Lower-load audit import (#218).** The importer now builds its de-duplication cache once per cycle and updates it in memory, instead of re-querying the database for every batch — cutting database load and memory use on big backlogs.
- **Faster Copilot activity import (#200, #223).** Adds indexes to the Copilot "accessed resource" lookup tables, and a **composite-key index on the accessed-resource junction** so the per-cycle de-duplication seeks instead of scanning. *(Database schema change — see below.)*
- **Usage-report import throttled to once a day without Redis (#211).** Previously the daily usage report could be re-imported on every cycle; it now runs at most once per day, cutting database load and Microsoft Graph calls.

### Reliability & data integrity
- **Blob-checkpoint data-loss fix (#210).** Fixes a case where a failed content download could let the checkpoint skip data, plus a race that could raise a duplicate-summary primary-key error.
- **Duplicate-key crash fix in concurrent-save audit import (#214).** With concurrent saves enabled, a duplicate lookup key could crash the batch save and abort the cycle; this is now handled so the cycle completes.
- **Correct de-duplication of Copilot accessed resources (#204, #205).** A volatile token in the SharePoint resource URL was preventing duplicate rows from collapsing; it is now stripped so the same resource de-duplicates correctly. The one-off clean-up migration was also rewritten to run in linear rather than quadratic time.
- **Manual DB-upgrade scripts now record themselves correctly (#224).** A hand-run manual upgrade script could finish without stamping the migration as applied when the database was already clean, so the app's Health page kept reporting the upgrade as pending; the scripts now always record completion.

### Security
- **Patched vulnerable NuGet packages (#225).** Updated **AngleSharp** (0.17.1 → 1.5.2) and **System.Security.Cryptography.Xml** (10.0.8 → 10.0.10) to clear the security advisories flagged by NuGet audit, and refreshed the matching assembly binding redirects. No functional change.

### Upgrade the database without the installer (#203)
- A new `-RunDbUpgrade` switch on `Deploy-AppServiceContent.ps1` applies pending database schema upgrades directly, so you no longer have to run the full installer just to bring the schema up to date.

### Database upgrade steps ⚠️
This release adds **five database migrations**. They apply in the order below via any one of: the installer (`AnalyticsInstaller.exe --initdb`), the new `-RunDbUpgrade` deploy switch (#203), or the per-migration **manual SQL scripts attached to the published release** (idempotent and safe to re-run). **Run them in a maintenance window with the importer stopped** — several index or rewrite tables that can be very large on a mature tenant.

**About the timings below:** they scale with your data volume (row counts in the audit and Copilot tables), so they are rough — *near-instant on a small or new database, up to the high end on a large tenant with years of Copilot history*. On **Enterprise / Azure SQL Database / Managed Instance** the index builds run `ONLINE` (non-blocking); on **Standard / Express / LocalDB** the equivalent step briefly locks the table. The scripts print live progress (`RAISERROR … WITH NOWAIT`) and the command timeout is disabled so a long build won't time out.

| # | Migration (PR) | What it does | Approx. time |
|---|---|---|---|
| 1 | `202607101200001_IndexCopilotAccessedResourceLookups` (#200) | Narrows 3 Copilot lookup columns from `nvarchar(max)` → `nvarchar(850)` (a table rewrite), trims any over-width values, then adds a non-clustered index on each | **The heaviest — minutes up to ~30 min+.** Shrinking a `max`/LOB column can't run `ONLINE` even on Enterprise, so this is an **offline table rewrite** that holds a schema lock; schedule downtime for it. |
| 2 | `202607130900001_DedupCopilotAccessedResourceSiteUrls` (#204, #205) | Collapses millions of near-duplicate Copilot `site_url` rows to one canonical row per site, re-points the junction, and deletes the duplicates (in small batches) | **The most variable — depends on how many duplicates exist.** No-op in seconds on a clean DB; **tens of minutes to a few hours** on a tenant heavily affected by the old volatile-token bug. Batched, so it never holds a large lock. |
| 3 | `202607141000001_IndexAuditEventsTimeStamp` (#207) | Adds indexes on `audit_events(time_stamp)` and `ignored_audit_events(processed_timestamp)` | **Minutes** — `audit_events` is often the largest table. `ONLINE` where supported, otherwise a brief lock. |
| 4 | `202607231000001_IndexSitesSiteId` (#221) | Adds a non-clustered index on `sites(site_id)` | **Seconds to ~a minute** — `sites` is a small table. |
| 5 | `202607231700001_CoverCopilotAccessedResourceDedup` (#223) | Adds the 6-column composite index used by the Copilot merge de-duplication | **Seconds to a few minutes** — depends on the junction table's size. |

**Rough total:** a few minutes on a small/new tenant; on a large, long-running tenant budget **~1–3 hours**, dominated by migration #1 (the column narrowing) and, if affected, migration #2 (the duplicate clean-up).

### Installer / config changes
- **Config schema bumped to 1.9.0** for the new opt-in Power Platform import (#219, #220). Re-save your configuration in the installer so the setting is persisted; existing config files remain compatible.
- The networking wizard's paired **Storage / Storage table** private-endpoint fields (see the private-deployment fix above) are also part of this release.

### Code maintenance & diagnostics
- **New save-timing traces (#222)** expose where each audit-import cycle spends its time (including the Copilot merge cost), so future performance issues are easier to diagnose from the logs.
- Assorted internal-only changes with no user-visible effect: a System.Text.Json binding-redirect fix in Common.Entities (#199), expanded Profiling test coverage and contributor docs (#208), a flaky-unit-test fix (#226), contributor-documentation updates (#227), logging-level / log-message tidy-ups, and test-data and tooling tweaks.

---

_Changes since the last stable release (build 1684 / #198): #199, #200, #201, #203, #204, #205, #207, #208, #210, #211, #213, #214, #218, #219, #220, #221, #222, #223, #224, #225, #226, #227._


sambetts:
## 🔍 Release critique

**Verdict: ship-ready as an RC.** The content is CI-green on the dev tip (`74094ca`: Tests + Release build + Secret scan all ✅) and the classic release break-points are clean. A few operational checks are worth doing before *publishing* (vs. leaving it as the draft RC).

### Strengths
- **Schema safety is solid.** All 3 migrations are registered in `Entities.csproj`, the `LatestId` guard test was bumped to `202607141000001_IndexAuditEventsTimeStamp`, and a full local test pass (627 passed, no `AutomaticDataLossException`) confirms the model-snapshot chain is intact. Each migration is edition-aware (online→offline), idempotent, and ships a `*.manual.sql` for hand-upgrades.
- **Risk-averse defaults.** The headline scale feature (concurrent saves) defaults to `AUDIT_MAX_CONCURRENT_SAVES=1` = today's exact strictly-serial behaviour; `>1` is opt-in and was verified to produce 0 duplicate rows. Nothing changes for existing tenants unless they choose to.
- **Right fixes bundled together.** #210 closes a genuine data-loss hole (a failed blob checkpointed as "done" and skipped forever) that only exists *because* of the new blob checkpoint, so shipping them in the same stable line is correct.

### Risks / things to watch (ranked)
1. **Three schema migrations on potentially huge tables — the real operational risk.** On non-Enterprise / non-Azure SQL editions the index builds run **offline** (table locked for the build), and the Copilot site-URL dedup is a **one-way `DELETE`**. Mitigations are in place and the notes call for a maintenance window with the importer stopped — but **time all three on the largest test tenant** before publishing, and confirm the dedup is acceptable as a non-reversible cleanup.
2. **The PR itself won't show a Tests check.** `tests.yml` triggers on `ready_for_review` / push, not on PR-open, so the green signal here is from the dev tip rather than this PR's merge SHA. Optionally toggle draft→ready to get a check on the exact SHA; the full Release build runs on merge to `main` regardless.
3. **`-RunDbUpgrade` ships as the no-installer upgrade path for the first time.** Low code risk (same `CheckDbUpgraded` logic, and `InstallerLogs.cs` was hardened so the Windows Event Log write no longer throws under the restricted App Service identity), but **smoke-test it against a non-prod App Service** — it downloads `ControlPanelApp.zip` from the release and needs the App Service identity to hold DB rights.
4. **In-memory fallbacks reset on restart** (blob checkpoint + once-a-day usage-report throttle, when no Azure Storage connection is configured). No data loss — just a one-cycle re-process of the overlap window after a restart. Worth noting it's storage-config-dependent.

### Notes-quality nits (optional)
- The `InstallerLogs` Event-Log hardening isn't mentioned in the body — small, but it's what keeps the App Service upgrade path from crashing; a half-line under the `-RunDbUpgrade` bullet would cover it.
- Consider adding a short **pre-publish checklist**: (a) confirm the 3 `*.manual.sql` scripts are attached to the draft "Stable build" release (per the robust-migration convention), (b) `-RunDbUpgrade` smoke test on a non-prod App Service, (c) migration timing captured on a large tenant.

*Critique generated from the `origin/main..origin/dev` diff (19 commits) and CI status; no code-blocking issues found.*

PR #260, opened 2026-08-19
Copilot: import the three Graph Microsoft 365 Copilot usage reports

sambetts:
# Copilot: import the three Graph Microsoft 365 Copilot usage reports

Closes milestone #6.

## Why

Today 100% of our Copilot data comes from the audit log. Microsoft explicitly states audit data "is not
intended to be used as the basis for Copilot usage reporting", and that metrics like prompt count or active
user count built on it "might not be consistent" with their official reports. So our numbers can disagree
with the Microsoft 365 admin centre and we have no defence. These three reports are the numbers customers
will compare us against.

This does **not** replace the audit-log Copilot import — the two answer different questions, and having both
is the point (see *Product angle* below).

## Scope

Imports all three reports, behind a new opt-in toggle:

| Report | Grain | Table |
|---|---|---|
| `getMicrosoft365CopilotUserCountSummary` | per period, per app | `copilot_user_count_log` |
| `getMicrosoft365CopilotUserCountTrend` | per day, per app | `copilot_user_count_log` |
| `getMicrosoft365CopilotUsageUserDetail` | per user, per date, per period | `copilot_usage_user_activity_log` |

Aggregates are imported first: they are a few thousand rows whatever the tenant size, need no per-user joins,
and are immune to the identity-concealment problem below, so they still work where the per-user report can't.

## Transport

Uses the **beta endpoints with `$format=application/json`**, over exactly the same plumbing as every other
Graph usage report here: `ManualGraphCallClient`, paged by `LoadAllPagesWithThrottleRetries`, sharing its
throttling and retry budget. No bespoke transport.

Rows are read as `JObject` rather than a fixed DTO, on purpose. The reports carry one property pair per
Copilot surface (`wordEnabledUsers` / `wordActiveUsers`) and Microsoft keeps adding surfaces — Edge,
Microsoft 365 Copilot and Copilot Chat work/web all arrived in one revision. Discovering them dynamically is
what makes a new Microsoft app become new *rows* rather than a new column, a new DTO property and a schema
migration on every customer database. Unrecognised prefixes still import with a generated name; known ones map
to the names the admin centre uses, so `powerPoint` stays "PowerPoint" rather than becoming "Power Point".

## Endpoint behaviours this had to handle

Each of these fails *silently* rather than loudly if ignored:

- **`version` is optional and defaults to `v1`.** Omit it and every prompt-count and active-usage-days value
  is simply absent — no error. `CopilotReportRequest` always sends it.
- **Valid `period` values differ by version**: v1 `D7/D30/D90/D180/ALL`, v2 `D7/`**`D28`**`/D90/D180/ALL`.
  Mixing them is a plain 400, so the pairing is validated before the call.
- **The JSON nests its numbers one level down** — `adoptionByProduct`, `adoptionByDate`, and
  `copilotActivityUserDetailsByPeriod` for the per-user report. One user object therefore yields one row per
  period, which is exactly the table's `(date, user, period)` key.
- **Licensed users only, global cloud only, ~48h latency.** Reflected in the docs, log messages and cadence.

**Microsoft has not published the beta JSON schema for report version 2.** Those fields are therefore looked
up under more than one plausible name and stay null when absent. Because absence is ambiguous — it can mean
"this is a v1 response" *or* "the names differ from the ones we look for" — a response carrying no version 2
values leaves those columns **untouched** rather than writing NULL over prompt counts a previous import
captured. Blanking good data is the more damaging of the two possible mistakes. This is the one place the v1.0
CSV endpoints would have been stricter, since their column names are documented.

## Identity concealment — the important behaviour

If the tenant enables *"Display concealed user, group and site names in all reports"*, Graph still returns
`200 OK` with one row per licensed user — just with the UPN and display name replaced by hashes.

Feeding those through the usual get-or-create-user path would create **one junk user per licensed account**
(200,000 on a large tenant), permanently polluting `users` and every report built on it, and producing joins
that are *wrong* rather than missing. So:

- the per-user report is **not imported at all** in that state;
- the reason is recorded in `copilot_usage_report_import_log`;
- the Health page degrades with the admin-actionable reason (including where to turn the setting off);
- the tenant-level counts are unaffected, and the audit-log Copilot import is not affected by that setting at
  all, so Copilot reporting keeps working from the audit source.

There are two layers, because syntax alone can never prove an identity belongs to the tenant:

1. A UPN shape check (rejects a bare hash, `hash@hash`, and a 32-character all-hex local part) — defence in
   depth. The hash length is matched exactly so a genuine account spellable in hex
   (`0123456789abcdef@contoso.com`) is not rejected.
2. **The actual boundary: the loader will not invent a user on an email domain this database holds no users
   for.** Unrecognised domains are skipped and counted, with a log line pointing at the user-metadata import.
   (An empty `users` table falls back to creating, since there is nothing to validate against yet.)

## Schema

Migration `202608190725064_AddCopilotUsageReports`. Three new, empty tables — `CREATE TABLE` is metadata-only,
so this is effectively instant at any tenant size, needs no maintenance window, and rewrites no existing
table. The only lock on existing data is a brief schema lock on `dbo.users` for the foreign key.

**`copilot_user_count_log` is narrow/tall by design** — one row per (report type, period, date, app) instead
of a column per app. See *Transport* above for why that matters.

Trend rows leave `report_period_days` NULL: a daily count is the same number whichever window asked for it,
and SQL Server treats NULLs as equal in a unique index, so a D7 refresh updates the rows a D180 backfill
created rather than duplicating every day.

The per-user table follows the `*_user_activity_log` convention and is keyed
`(date, user_id, report_period_days)` — D7 and D28 describe the same user and date with *different* prompt
counts, active-day counts and last-activity values, so they're different facts, not a conflict.

Both indexes are UNIQUE and **functional, not performance-motivated**: they're the keys the importer upserts
on, which is what stops Graph's 3-day rolling gap-fill from duplicating rows on re-import. Per the repo rule
that a performance index must ship with a measured before/after benchmark, no covering index is included here
— that should be added *and measured* when a report query against these tables actually exists.

All text columns are `nvarchar`, so localised app names and non-Latin display names round-trip.

`<migrationid>.manual.sql` is included and was validated end to end: applies from the predecessor state,
stamps `__MigrationHistory` with a `Model` **byte-identical** to the `.resx` snapshot, EF then reports no
pending migrations, and a re-run is a no-op. Because this migration changes the EF model it cannot copy the
predecessor's `Model` blob (EF would then try to create the tables again), so the correct snapshot is embedded
as chunked base64; `ProductVersion` is stamped explicitly as `6.5.2` to match an installer-driven upgrade.

## Cadence

Copilot reports get their own `GraphCopilotUsageReportsIntervalHours`, defaulting to **24h on every preset**.

They deliberately do *not* reuse `GraphMetadataImportIntervalHours`: that is preset-derived and **0 (= every
cycle) on the default `High` preset**, with a 10-minute cycle pause. Reusing it would have re-downloaded and
re-processed every licensed user's Copilot row every 10 minutes against a source that only refreshes every
~48 hours. This feature has no legacy every-cycle behaviour to preserve. `ForceGraphMetadataImport` still
overrides it for one run. Runtime AppSetting, not installer config schema, so no further `CONFIG_VERSION` bump.

## Behaviour at ~200,000 users

- Saving works in **batches of users**: each batch queries only its own existing rows, saves, and detaches.
  That bounds EF's change tracker as well as the SQL command trees — the tracker, not the batch size, is what
  actually blows up at 200k rows.
- Users are resolved with **one** query into a case-insensitive dictionary, never per row. No `.ToLower()` in
  any EF predicate (non-SARGable, forces a scan). New users are flushed **and detached** per batch, so
  `DetectChanges` stays O(batch) rather than O(total × batches).
- Unchanged rows are dirty-checked and skipped. The report refresh date deliberately does **not** count as a
  change on its own — it advances daily, so including it would rewrite the whole window (up to 180 days × every
  app) every day purely to restamp provenance. Report-level freshness lives in the import log.
- `AutoThrottleHttpClient` only enforced `MaxRetries` on the *no-`Retry-After`* path. Graph almost always sends
  that header, so a persistently throttled endpoint retried **forever** — a hang, not an exception, which no
  caller-side handling can recover from. Pre-existing, but this adds a new heavy Graph caller and it directly
  undermined the failure-isolation claim below. The budget is now enforced on every 429.

### Known limitation: group-filtered tenants

The configured Entra group filter applies to the per-user report, and is one `/memberOf` call per user (cached
for an hour) — what the ten existing per-user usage-report loaders already do. It short-circuits entirely when
no filter is configured, the common case. A group-filtered 200k-user tenant will find this slow, so the loader
logs the number of lookups it is about to make rather than quietly taking hours. Bulk group resolution is a
pre-existing cross-cutting problem for all of those loaders and is deliberately not solved here.

A related pre-existing issue worth a separate look: `UserGroupsCache` treats a *failed* membership lookup as
"in scope", so a Graph blip can import users outside the configured group. Shared code, not changed here.

## Failure isolation

- each report runs independently, in **its own `DbContext`**, and logs instead of throwing;
- the section reports success/failure, and the cadence key is only stamped on success, so it retries next
  cycle. `FinishedSectionImport` is only emitted on success, so a permanently failing section stops reporting
  itself as healthy;
- a failure no longer unwinds out of `GetAndSaveAllGraphData` and skips the Teams and sent-email imports that
  run after it. The backfill-status query is also non-throwing for the same reason;
- the "have we done the one-off D180 backfill?" decision is based on a **completed D180 trend import** in the
  import log, not "any Copilot row exists". The latter meant a successful summary alongside a failed D180
  trend permanently downgraded every later run to D28 and lost the backfill for good;
- download, parse **and persistence** failures are all recorded in `copilot_usage_report_import_log`, written
  on a fresh context so a poisoned one can't swallow the diagnostic;
- **a response we can't parse fails loudly** instead of looking like "this tenant has no Copilot licences".
  If Graph returns entries but none yield rows, the import throws rather than recording an empty snapshot —
  which would otherwise mark the D180 backfill complete and lose the history.

## Config / installer

New `ImportTaskSettings.GraphCopilotUsageReports` (opt-in, default false), a checkbox on the installer's
Import Targets page, and `CONFIG_VERSION` bumped **1.9.0 → 1.10.0** with a `// History:` entry, in this same
change. `SolutionInstallVerifier` now verifies the Graph usage-report permission when *either* usage-report
toggle is on.

No new Graph permission: `Reports.Read.All` is already in the base solution permission set.

## Health page

`api/Health/data` now reports, per Copilot report: last import time, whether identities were concealed, and
any error. Previously a failed download would have looked like a recent healthy import.

## Tests

24 tests in `CopilotUsageReportImportTests`: v1/v2 parsing (v2-only values land as NULL, not 0), new apps
appearing with no code change (including one Microsoft hasn't shipped), app-name mapping, trend vs summary
grain, one row per report period, non-Latin app names, `version` always on the URL, wrong-period-for-version
rejection, concealed and email-shaped-pseudonym rejection, refusal to invent a user on an unrecognised domain,
unparseable-shape failure, upsert idempotency, gap-fill revision, refresh-date-only no-op, D7/D28 coexistence,
and that a response with no v2 values doesn't blank stored v2 data.

`UrlFullUrlMigrationPipelineTests.LatestId` updated. Its `ConnStr()` helper also had to stop constructing a
`DbContext`: those tests deliberately park the DB on an old migration, and in Release the
`CreateDatabaseIfNotExists` initializer validates the model against the last `__MigrationHistory` row. That
stayed quiet only while every recent migration was raw-SQL reusing its predecessor's snapshot — this is the
first model-changing migration since `RemoveDataverseTables`.

Full solution builds clean. 155 passing across the affected suites; the only failures are tests that construct
`AppConfig`, which needs the CI-injected `__TenantGUID__` secret and fails locally on any branch.

Reviewed across three independent passes on different models; findings on failure isolation, the per-user key,
change-tracker growth, the concealment boundary, the throttle budget and the import cadence are all addressed.

## Product angle worth stating

These reports cover **licensed users only** — Microsoft excludes unlicensed Copilot Chat usage from the
reports APIs entirely. Our audit-log import *does* see unlicensed Copilot Chat. Combining the two lets us show
**"unlicensed users actively using Copilot Chat"** — a licence-demand / upsell list that Microsoft's own
reports cannot produce.

The per-user report also gives us the first real adoption funnel:
Licensed → Enabled → Ever active → Active in last 28 days → Daily active, plus `active_usage_days` (habit, not
just "touched it once") and the Copilot agent last-activity date, the only agent signal in any Graph report.

## Deferred

**Power BI (`reports/Copilot/*.pbit`) is not included** — tracked in #261. A `.pbit` is a binary package that
can't be meaningfully validated without Power BI Desktop, and a malformed template fails to open for
customers. The issue carries the funnel/licence-waste spec and the exact tables to build against.

## Reviewer hotspots

1. `CopilotUsageUserDetailLoader` — the concealment abort path and the bulk user resolution.
2. The NULL-`report_period_days` trick for trend rows (verified: SQL Server's unique index does reject a second
   `(Trend, NULL, date, app)` row).
3. The version-2 field-name guesses and the "don't blank on absence" rule that compensates for them.
4. `GraphImporter.HasCompletedTrendBackfill` — the one-shot D180 backfill must be genuinely self-healing.
5. The embedded model blob in the manual SQL script.


sambetts:
Power BI follow-up tracked in #261 (same milestone) - the `.pbit` files are binary Power BI templates that can't be authored or validated outside Power BI Desktop, so that work is deliberately not in this PR.

sambetts:
Pushed `63e1e92` after a further independent review pass on two other models (GPT-5.6 Sol and Gemini 3.1 Pro). One finding was material enough to call out explicitly for reviewers:

**Cadence — the section would have run every 10 minutes on a default install.**

It reused `GraphMetadataImportIntervalHours`, which is preset-derived and **0 (= every cycle) on the `High` preset**, with a 10-minute cycle pause. So a default installation would have re-downloaded and re-processed every licensed user's Copilot row every 10 minutes, against a source that only refreshes roughly every 48 hours — while the code comment claimed a daily cadence.

Copilot reports now have their own `GraphCopilotUsageReportsIntervalHours` setting, defaulting to **24h on every preset**. This feature has no legacy every-cycle behaviour to preserve, so it doesn't inherit the `High` preset's 0. `ForceGraphMetadataImport` still overrides it for one run. This is a runtime AppSetting on `AppConfig`, not the installer config schema, so no further `CONFIG_VERSION` bump.

Also in this commit:

- **Throttle budget.** `AutoThrottleHttpClient` only enforced `MaxRetries` on the *no-`Retry-After`* path. Graph almost always sends that header, so a persistently throttled endpoint retried **forever** — a hang, not an exception, which no caller-side handling can recover from. Pre-existing, but this feature adds a new heavy Graph caller, and it directly undermined the failure-isolation claim in the description. Budget is now enforced on every 429, and the terminal response is disposed.
- **v2 detection was too weak** — it required only one of two headers, so a response missing some v2 columns was treated as v2 and could write NULL over prompt counts a previous good import had stored. Now requires the full v2 header set.
- **Partial aggregate parses are recorded as incomplete**, so they stay retryable. This mattered: the one-off D180 history backfill could otherwise be marked complete from a partial parse, switching later runs to D28 and losing the missing history permanently.
- **Memory**: dropped the parsed-but-never-persisted display name and a full-size list copy in the per-user save path.

Both reviewers independently confirmed the two design decisions I most wanted challenged: the NULL `report_period_days` trick for trend rows is correct (SQL Server's UNIQUE constraint permits only one NULL per key combination), and neither could find a bypass of the concealed-identity boundary. One verified the manual script's embedded base64 model byte-for-byte against the `.resx`.

Tests: 22 Copilot tests pass. The 38 failures in the wider local run are all the same `__TenantGUID__` CI-secret placeholder and are unrelated to this branch.

sambetts:
Switched to the beta JSON endpoints per review feedback (`4ef69b7`). The v1.0 CSV transport and everything built for it - `CsvReportTable`, the streaming reader, manual redirect handling, and the non-redirecting handler overload on `ConfidentialClientApplicationHttpHandler` - are gone; that shared class is back to how it was. Net **-119 lines**, and the import now runs on exactly the same plumbing as every other Graph usage report here.

Rows are read as `JObject` rather than a fixed DTO so the narrow/tall benefit survives the transport change: a new Microsoft Copilot surface still becomes new *rows*, not a new column plus a migration. There's a test for an app Microsoft hasn't shipped yet.

One deliberate trade-off worth flagging: Microsoft hasn't published the beta JSON schema for report **version 2**, so those field names are best-effort (looked up under a couple of plausible names, null when absent). The CSV endpoints documented their column names, so that path could fail hard on a v1-shaped response. Since absence is now ambiguous - v1 response, or names we don't recognise - a response with no v2 values leaves those columns **untouched** instead of writing NULL over prompt counts a previous import captured. Blanking good data is the worse of the two mistakes. Covered by `UserDetailLoader_DoesNotBlankVersion2DataWhenAResponseCarriesNone`.

Description updated. 24 Copilot tests pass; schema, migration and manual SQL are unchanged - only the transport moved.

sambetts:
## Blocked: the migration snapshot was scaffolded without #262

Validated locally on a scratch merge. Same 26 migration tests, same filter, each after dropping the LocalDB test database and a clean rebuild:

| Branch | Migration tests |
| --- | --- |
| plain `dev` | **26 passed** |
| `dev` + #262 | **38 passed** |
| `dev` + #262 + **this PR** | **18 failed**, 8 passed |

Failure: `There is already an object named 'copilot_event_accessed_resource_actions' in the database.`

### Root cause

That table is created only by #262's migration. The duplicate comes from EF auto-migrating, not from this PR's explicit SQL.

The `.resx` model snapshots give it away:

| Migration | Snapshot size |
| --- | --- |
| `202608131055001_IndexReportDateQueries` (dev) | 51,364 |
| `202608190622001_CopilotDroppedAuditFields` (#262) | 53,852 |
| `202608190725064_AddCopilotUsageReports` (this PR) | **53,548** |

This PR's migration ID sorts **last**, so EF compares the live model against *its* snapshot — but that snapshot is **smaller** than #262's, proving it was scaffolded against `dev` before #262 existed. EF sees #262's entities missing from the latest snapshot, and because `Migrations/Configuration.cs` has `AutomaticMigrationsEnabled = true` it generates an on-the-fly migration that re-creates #262's tables.

This is the trap documented in `src/AnalyticsEngine/.github/copilot-instructions.md` → *"`AutomaticMigrationsEnabled = true` + an outdated snapshot"*.

### Fix

**#262 is now merged into `dev`.** Re-scaffold this PR's migration on top of it — the canonical way per the same instructions:

> `Add-Migration -Force <Name>` in the Visual Studio Package Manager Console (with `Entities` as the project, `WebJob.Office365ActivityImporter` as the startup project).

So: merge latest `dev` into this branch, then `Add-Migration -Force AddCopilotUsageReports`, and confirm the regenerated snapshot is **larger** than #262's. Also regenerate `202608190725064_AddCopilotUsageReports.manual.sql` if the scaffold changes the SQL.

### Conflicts to expect when merging `dev` in

Already resolved once on the scratch branch, for reference:

- `Common/Entities/Config/AppConfig.cs` — union: keep both `SentEmailNoMailboxRetryHours` (#263) and `GraphCopilotUsageReportsIntervalHours` (this PR).
- `Common/Entities/Entities.csproj` — union: register **both** migrations' `.cs`, `.Designer.cs` and `.resx` entries.
- `Tests.UnitTests/UrlFullUrlMigrationPipelineTests.cs` — `LatestId` must become `"202608190725064_AddCopilotUsageReports"` (this PR's, as the later timestamp).

No changes needed to the feature code itself — this is purely the EF snapshot chain.

Issue #271, opened 2026-08-19
CI: detect EF migration snapshot conflicts between PRs before they reach dev


sambetts:
## Problem

Two PRs can each add an EF6 migration that is valid **in isolation**, and still break `dev` the moment both are merged — with nothing in CI to catch it.

Each migration's `.resx` holds a model snapshot. If PR B is scaffolded before PR A merges, and B's migration ID sorts **after** A's, then once both land the newest snapshot in the chain does not know about A's entities. Because `Migrations/Configuration.cs` sets `AutomaticMigrationsEnabled = true`, EF then generates an on-the-fly migration to reconcile the difference — which either re-creates A's objects or, for removals, throws `AutomaticDataLossException`.

## Real example

#262 and #260 were each green on their own branch. Validated locally on a scratch merge, same 26 tests, same filter, each after dropping the LocalDB test database and rebuilding:

| Branch | Migration tests |
| --- | --- |
| `dev` | 26 passed |
| `dev` + #262 | 38 passed |
| `dev` + #262 + #260 | **18 failed** |

Failure: `There is already an object named 'copilot_event_accessed_resource_actions' in the database.` — a table created only by #262's migration.

The snapshot sizes made it provable: #260's migration sorts last but its snapshot is **53,548 bytes against #262's 53,852**. Smaller despite being later, i.e. scaffolded before #262 existed.

#260 was held back rather than merged. Nothing automated would have stopped it.

## Proposed fix

A CI job that, for any PR touching `Common/Entities/Migrations/`:

1. merges the PR into the **target branch** (not just builds the PR head);
2. drops/creates a clean database;
3. runs the migration test suite against it.

That is the only configuration in which this class of failure appears, and it is cheap to run because it is gated on the migrations path.

A useful cheap pre-check alongside it: assert the newest migration's snapshot is not smaller than its predecessor's, which is what exposed this instance.

## Acceptance criteria

- A PR whose migration snapshot predates a migration already on the target branch fails CI.
- The check only runs when `Migrations/` is touched.
- The failure message points at the snapshot chain rather than just the SQL error.

sambetts:
Implemented in #282 (merged to dev): MigrationSnapshotTests.cs fails the build when a migration snapshot predates the target branch. Shipping in release PR #283.

PR #292, opened 2026-08-20
Copilot licence adoption tool: find unused seats and unlicensed heavy users

sambetts:
Adds a **Copilot Adoption** area to the web app: a licence-decision tool built on data the product already imports. No schema changes, no migrations, no config-schema change.

Milestone: **Copilot licence adoption tool** (#12). First cut — expect review comments.

## Why

Two questions decide Copilot spend, and neither is answerable today:

1. **Who holds a Copilot licence and isn't getting value from it?** Microsoft's report gives "active users", which is a yes/no. Someone who opened Copilot once and someone who lives in it produce the same count and need opposite responses (reclaim the seat vs. recruit them as an advocate).
2. **Who's a heavy Microsoft 365 user without a licence?** Including — the part nothing else can see — people **already using Copilot Chat with no seat**. Microsoft's reports APIs exclude unlicensed users entirely; our audit-log import doesn't.

## Architecture

| Layer | Location | Responsibility |
|---|---|---|
| Analysis engine | `Common/Entities/CopilotAdoption/` | SKU classification, scoring, SQL builders, CSV, orchestration |
| API | `Web/Controllers/CopilotAdoptionAPIController.cs` | Caching, filtering, paging, CSV responses |
| UI | `Web/Scripts/admin-app/src/{pages,components/copilotAdoption}` | Overview / Licensed users / Opportunities / methodology |

**The engine deliberately lives in `Common.Entities`, not the web project.** The stated requirement was that this be reusable for automated reports later; `CopilotAdoptionService.AnalyseAsync()` returns the complete scored dataset and has no web dependency, so a web-job can call it unchanged. It also means every rule is unit-testable without a controller.

### Files

**`CopilotLicenceClassifier.cs`** — decides which `license_types` rows are a Microsoft 365 Copilot *seat*. Ordered rules: exclusion list → SKU part-number prefix → conservative display-name fallback. Prefix matching (not an exact list) because a SKU newer than the shipped licensing CSV has **no display name** — the importer stores the part number as the name — so an exact-list classifier would silently miss a customer's newest and most expensive seats. Copilot Studio / Copilot for Sales / Sales Copilot are explicitly excluded; counting them would inflate the denominator and understate adoption, which is the failure mode that matters when the number is used to argue about spend. The classification for **every** licence type is returned to the UI and can be overridden per request, so a misclassified SKU is a visible, correctable fact rather than a wrong headline.

**`CopilotAdoptionScoring.cs`** — all the maths, as pure functions. Engagement = frequency (50%) + depth (30%) + breadth (20%), each capped at its target:

- *Frequency*: active days vs. 60% of **working** days in the window. Calendar days would cap a genuinely daily user at ~71%.
- *Depth*: interactions per active day vs. 5. Separates "opened it" from "worked with it".
- *Breadth*: distinct surfaces vs. 3. Single-surface users are the cheapest group to move.

Bands: Champion 75+, Established 50+, Developing 25+, Trialling >0. Zero-activity users are split **before** thresholds into `Dormant` (used it within `HistoryDays`) and `NeverUsed`. Every threshold and weight is in `CopilotAdoptionOptions` and echoed back in the API response, so a figure can be traced to the rule that produced it.

Opportunity scoring weights unlicensed Copilot Chat use at 35 vs. 25/20/20 for Teams/email/documents — evidence over inference.

**`CopilotAdoptionSql.cs`** — every query, as strings (testable, and shown in the existing SQL popover).

**`CsvSerialiser.cs`** — RFC 4180, UTF-8 **with BOM**, formula-injection defused, culture-invariant.

**`CopilotAdoptionExports.cs`** — filter/sort/page + the two CSV column schemas.

## Performance (target: ~200k-user tenant)

- Licensed population from `user_license_type_lookups` filtered by `license_type_id` — index-only seek on the leading key of `IX_license_type_id_user_id`, **not** a scan of `users`.
- Copilot audit history read **once** over a bounded range; the window-vs-history split is `CASE` inside a single aggregate rather than two joins. This is the most expensive join in the product.
- M365 figures from **one** resolved snapshot date per workload → equality seek on `IX_date` (per `IndexUsageReportSnapshots`), not a range scan.
- Unlicensed candidates driven **from the activity tables**, anti-joined against the licensed set. Driving from `dbo.users` would scan every account to discard the ~majority with no activity.
- `OPTION (RECOMPILE)` throughout; the `copilot_chats`→`audit_events` join is left **un-hinted**, per the measurements recorded on `ReportsAPIController.CopilotAuditNaturalJoin` (a forced MERGE join degraded to a flat 73,857-read full scan at every window).
- Analysis runs once per (window, licence selection) and is cached for 10 min; the cached value is the `Task`, so concurrent first-hits share one execution rather than each starting a full audit scan.
- Licensed users are scored **in memory** (capped at 50k, warns if hit) so the scoring rules have exactly one implementation. Copilot seats are bought individually, so this is far below any real tenant. Unlicensed candidates can't be — that population is the whole directory — so they're ranked in SQL by an expression generated from the *same* options object, then re-scored in C# before display.

**No schema change.** `Migrations/`, `Create DB.sql` and `CONFIG_VERSION` are untouched — verify with `git diff origin/dev --stat -- src/AnalyticsEngine/Common/Entities/Migrations src/AnalyticsEngine/Common/Entities/Resources`.

## Correctness details worth a look

- **Per-user audit fallback.** If the audit import is healthy overall but has nothing for a given user while Microsoft's report shows them active, the report wins for that user. Without this, a lagging or partly-failed import puts actively-working people on a list of seats to take away. The source used is recorded per row (`signalSource`) and exported.
- **Cowork is not reported as "0%" when undetected.** On a tenant not enabled for Cowork, `coworkDetected: false` suppresses the metric — "0% Cowork adoption" reads as a failure and would be quoted as one.
- **Disabled accounts** are excluded from candidates (can't use a licence) but **kept** in the licensed list — a disabled account holding a seat is the clearest reclaim there is. The `disabledOnly` filter treats `null` (not imported) as *not* disabled.
- **Segments below 5 seats are dropped** from the department/country charts; "0% of 2 seats" at the top of an executive chart invites the wrong call. Absolute counts are shown alongside every percentage.
- **Empty licence list** yields `IN (-1)`, not `IN ()` — a tenant with no Copilot licences gets an empty report, not a 500. That's a legitimate use ("should we buy Copilot?"), and the opportunity side still works.
- **Sorting always tie-breaks on UPN** so paging can't duplicate or drop a user between a page and an export.

## Tests

**62 new tests, all passing.** Pure tests cover SKU classification, scoring/banding, opportunity ranking, CSV escaping/Unicode/injection/culture, filtering, sorting, paging, summary assembly and controller parameter handling.

`CopilotAdoptionSqlIntegrationTests` runs the **real SQL** against a throwaway database (`ScratchDatabase`, same pattern as the migration tests) via `Database.SqlQuery<T>`, so DTO materialisation is exercised too. That was worth doing — it caught three defects nothing else would have:

| Defect | Impact if shipped |
|---|---|
| `Read` used as an unbracketed alias | `Incorrect syntax near the keyword 'Read'` — the entire opportunities list 500s |
| Ranking expression referenced CTEs omitted when a source is unavailable | `The multi-part identifier "teams.Messages" could not be bound` on any audit-only deployment |
| Per-user audit gap classified active users as `NeverUsed` | Active people listed as seats to reclaim |

A Greek-alphabet round-trip test guards the Unicode requirement end-to-end (a `varchar` anywhere on the path would mangle it in a document about to go to an executive).

```
Total tests: 62   Passed: 62
```

### Pre-existing failures, not from this PR

`MigrationSnapshotTests.LatestMigrationSnapshot_IsUpToDateWithTheEntityModel` fails, which cascades into `EntityTests` (`AutomaticDataLossException`). Verified by stashing this branch and running it against pristine `origin/dev` — it fails identically there. The pending op is `DropForeignKey("dbo.copilot_interactions", ...)`, from the AI-interaction-history entities already on `dev` (issue #271). This PR adds no entities and no migrations.

## Reviewer hotspots

1. **`CopilotLicenceClassifier` prefix/exclusion lists** — the highest-consequence judgement here. A missed seat SKU understates the denominator.
2. **The weights and thresholds in `CopilotAdoptionOptions`** — defensible defaults, but they're opinions. Worth a second view given the audience.
3. **`MaxLicensedUsersScored = 50000`** — in-memory scoring cap. I believe it's comfortably above any real Copilot seat count; if you disagree the scoring would need to move into SQL, which costs the single-implementation guarantee.
4. **`CopilotAdoptionSql.LicenceOpportunitiesSql`** — the only query whose cost is bounded by activity rather than by seats. Ranking is TOP 5,000.
5. **Cowork detection** — `app_host = 'cowork'` OR a `Copilot.M365Copilot.Cowork*` agent id. Please sanity-check against real payload shapes.

## Not in this PR

- Scheduled/emailed reports. The engine is shaped for it (`AnalyseAsync` → complete dataset, no web deps) but no job is wired up.
- Seat cost input, so "reclaimable seats" can't be shown as a currency figure. Deliberate: it needs a config-schema change and a `CONFIG_VERSION` bump.
- Wiki documentation — happy to write it once the shape is agreed.
- No benchmark table: this PR adds no indexes and no schema changes, so the *Prove every schema change improves performance* gate doesn't apply. The query shapes follow the patterns already measured for the Reports area.

### DIFF
--- a/src/AnalyticsEngine/Common/Entities/CopilotAdoption/CopilotAdoptionExports.cs
+++ b/src/AnalyticsEngine/Common/Entities/CopilotAdoption/CopilotAdoptionExports.cs
@@ -1,0 +1,370 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Common.Entities.CopilotAdoption
+{
+    /// <summary>
+    /// How the licensed-user list is filtered and ordered. A plain object rather than a set of method
+    /// arguments so the same shape can come from a query string today and from a scheduled report's
+    /// configuration later without changing any of the logic below.
+    /// </summary>
+    public class LicensedUserQuery
+    {
+        /// <summary>Free-text match against UPN, mail, department, job title or manager.</summary>
+        public string Search { get; set; }
+
+        /// <summary>Restrict to these engagement bands; empty means all.</summary>
+        public List<AdoptionBand> Bands { get; set; } = new List<AdoptionBand>();
+
+        public string Department { get; set; }
+
+        public string Country { get; set; }
+
+        /// <summary>Only users who have used Microsoft 365 Copilot Cowork.</summary>
+        public bool CoworkOnly { get; set; }
+
+        /// <summary>Only users whose account is disabled - the clearest reclaim candidates of all.</summary>
+        public bool DisabledAccountsOnly { get; set; }
+
+        public double? MinScore { get; set; }
+
+        public double? MaxScore { get; set; }
+
+        /// <summary>One of <see cref="LicensedUserSortFields"/>.</summary>
+        public string SortBy { get; set; } = LicensedUserSortFields.Score;
+
+        /// <summary>
+        /// Ascending by default for the score, because the whole point of the list is finding the
+        /// people who are <i>not</i> using their licence.
+        /// </summary>
+        public bool SortDescending { get; set; }
+    }
+
+    /// <summary>Sortable columns of the licensed-user list. An allow-list: nothing here reaches SQL.</summary>
+    public static class LicensedUserSortFields
+    {
+        public const string Score = "score";
+        public const string UserPrincipalName = "upn";
+        public const string Interactions = "interactions";
+        public const string ActiveDays = "activeDays";
+        public const string LastUse = "lastUse";
+        public const string Department = "department";
+        public const string Cowork = "cowork";
+    }
+
+    /// <summary>How the licence-opportunity list is filtered and ordered.</summary>
+    public class LicenceOpportunityQuery
+    {
+        public string Search { get; set; }
+
+        public string Department { get; set; }
+
+        public string Country { get; set; }
+
+        /// <summary>Only candidates that clear the recommendation threshold.</summary>
+        public bool RecommendedOnly { get; set; }
+
+        /// <summary>Only candidates who are already using Copilot Chat without a licence.</summary>
+        public bool ExistingCopilotUsersOnly { get; set; }
+
+        public double? MinScore { get; set; }
+
+        /// <summary>One of <see cref="LicenceOpportunitySortFields"/>.</summary>
+        public string SortBy { get; set; } = LicenceOpportunitySortFields.Score;
+
+        /// <summary>Descending by default: the strongest business cases first.</summary>
+        public bool SortDescending { get; set; } = true;
+    }
+
+    /// <summary>Sortable columns of the licence-opportunity list.</summary>
+    public static class LicenceOpportunitySortFields
+    {
+        public const string Score = "score";
+        public const string UserPrincipalName = "upn";
+        public const string CopilotUse = "copilot";
+        public const string Collaboration = "collaboration";
+        public const string Email = "email";
+        public const string Documents = "documents";
+        public const string Department = "department";
+    }
+
+    /// <summary>
+    /// Filtering, sorting and paging over an analysis result, plus the CSV export schemas.
+    ///
+    /// All of it operates on the already-scored, already-materialised lists rather than going back to
+    /// the database, which is what guarantees the exported CSV is exactly the data behind the summary
+    /// the user is looking at. Nothing is more damaging to a report used in a licence negotiation than
+    /// a spreadsheet that disagrees with the chart it was exported from.
+    /// </summary>
+    public static class CopilotAdoptionExports
+    {
+        #region Licensed users
+
+        /// <summary>Applies the filter, then the sort. Returns a new list; the source is not mutated.</summary>
+        public static List<LicensedUserAdoptionRow> Apply(
+            IEnumerable<LicensedUserAdoptionRow> rows,
+            LicensedUserQuery query)
+        {
+            var source = rows ?? Enumerable.Empty<LicensedUserAdoptionRow>();
+            var q = query ?? new LicensedUserQuery();
+
+            var filtered = source.Where(row => MatchesLicensedUser(row, q));
+            return SortLicensedUsers(filtered, q).ToList();
+        }
+
+        private static bool MatchesLicensedUser(LicensedUserAdoptionRow row, LicensedUserQuery q)
+        {
+            if (q.Bands != null && q.Bands.Count > 0 && !q.Bands.Contains(row.Band))
+            {
+                return false;
+            }
+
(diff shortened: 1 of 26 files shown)


sambetts:
### Follow-up: the tab was invisible on a developer machine (27579bf)

The tab was correct in `App.tsx` all along, and CI would have shipped it fine. It didn't appear locally because **the build silently skipped rebuilding the SPA**.

`Web.csproj`'s `NpmInstall` target did three jobs — `npm install`, `npm run build`, and registering the built assets as `Content` — but its incremental check declared only `Scripts/admin-app/package.json` as an input, against a stamp file in `node_modules` as the output. `package.json` changes only when a *dependency* changes, so once a machine had built once the stamp was permanently newer and MSBuild skipped the whole target:

```
stamp        : 08/17/2026 09:14:11   <- output
package.json : 07/02/2026 07:34:41   <- only declared input
```

Editing any page or component left the previously-built bundle in place, with nothing in the build log to say why. CI was unaffected — a fresh checkout has no `node_modules`, so the target always ran. That's why this survived until a change was big enough for someone to go looking for it.

The skip had a second effect worth noting: the `Content` items were declared *inside* the same target, so a skipped run also failed to register the built bundle — it wasn't copied into the build output or the publish package either.

**Fix** — split into three targets with correct inputs:

| Target | Inputs | Outputs |
|---|---|---|
| `NpmInstall` | `package.json`, `package-lock.json` | `node_modules/.install-stamp` |
| `NpmBuildSpa` | `src/**`, `public/**`, `index.html`, `vite.config.ts`, `tsconfig.json`, `package.json` | `build/index.html` |
| `IncludeSpaBuildOutput` | *(always runs)* | registers `build/**` as `Content` |

`IncludeSpaBuildOutput` can't be a top-level `ItemGroup` — on a first build those files don't exist at project-evaluation time, which is why it was inside a target originally.

**Verified:**
- fresh build → installs + builds
- `touch src/App.tsx` → `NpmBuildSpa` re-runs (this is the case that was broken)
- no-change rebuild → both npm targets skip, `Content` still registered
- CI-style publish → `_PublishedWebsites/Web` contains the bundle, with `Copilot Adoption` present in `index-*.js`

**To see the tab on an existing checkout**, either rebuild the Web project (now sufficient), or:

```
cd src/AnalyticsEngine/Web/Scripts/admin-app
npm run build
```

PR #315, opened 2026-08-23
Copilot Adoption: explainability, AI-in-One capability parity, Excel export, correctness fixes and persona demo data

sambetts:
## Summary

Follow-up on the first cut of the Copilot Adoption area (`/copilot-adoption`). Five changes, one of them a straight bug fix, plus a new wiki specification page that future changes to this area should be reviewed against.

Wiki page (already pushed): [Copilot Adoption Tool](https://github.com/pnp/Microsoft365-Analytics-Insights/wiki/Copilot-Adoption-Tool)

---

## 1. Bug: the methodology tab rendered `NaN%` and blank thresholds

`CopilotAdoptionOptions` was the **only** model in `Common/Entities/CopilotAdoption/` without `[JsonProperty]` camelCase attributes, and this solution has no global camelCase contract resolver — so it serialised as PascalCase while the TypeScript client read camelCase. Every threshold and weight came back `undefined`, producing:

> **Frequency (NaN% of the score).** … against a target of NaN% of the available working days (assuming working days a week).
> **Bands.** Champion at +, Established at +, Developing at +, Trialling below that.

All 30 properties now carry explicit camelCase names. `EveryTuningValue_IsSerialisedInCamelCase` asserts every serialised property starts lower-case and that the count matches the reflected property count (less the static `Default`), so a newly added option that forgets the attribute fails the build rather than silently blanking the page.

## 2. Every assertion explains itself, on the spot

New `InfoTip` component — a small "i" in the top-right of each KPI card, chart card header and table column header. Each answers four things: **what** the figure claims, **how** it is worked out, the **formula** verbatim, and the **source** import with its caveats.

Applied to all eight headline figures, every overview chart, and the `Engagement`, `Band`, `Active days`, `Action`, `Business case`, `Already using Copilot` and `Justification` columns. Deliberately *not* applied to self-defining figures such as adoption by country — the rule recorded in the wiki is "if the tool applied a threshold, a weight, a normalisation or a classification, it must explain itself".

## 3. The repeated "Recommended action" column

The licensed-user list printed an identical paragraph on every row of a band — a hundred copies of one sentence presented as a hundred findings.

`CopilotAdoptionScoring` now emits `RecommendedActionCode` + `RecommendedActionLabel` alongside the existing prose, driven by the same branching. On screen each row carries a two-word tag, the meaning is stated once in a collapsible legend above the table (showing only the actions present in the current filter), and a new **Enablement plan** card on the Overview aggregates them into how many people need each job — which is the form an admin can actually plan from.

The CSV keeps the full sentence on every row and gains a short-label column, because a spreadsheet gets sorted and filtered and cannot rely on a nearby legend.

While wiring the tag up, the Established/low-breadth prose was found to open with "Sustain and broaden" while its tag said "Broaden". Reworded, and `TheActionTagAndTheExportedProse_NeverDisagree` now pins every band/breadth combination.

## 4. The "business case" score

Column-header tip on the opportunities list gives the full weighting (35 / 25 / 20 / 20), each component's target, the recommendation threshold, and why proven unlicensed Copilot use is weighted heaviest. The methodology tab gains the verbatim formula and an explanation of why each component is capped before weighting.

## 5. New visuals, from `microsoft/AI-in-One-Dashboard`

| Visual | What it adds |
|---|---|
| **Habit formation** tiles | Active users bucketed Infrequent / Moderate / Frequent / Daily by active days — the unweighted counterpart to the engagement bands |
| **Engagement mix** donut | Replaces the band bar chart; the message is the split of a whole, not a comparison of categories |
| **Usage by app** treemap | Replaces the bar chart; a top surface with 10x the traffic reads as an area, whereas as a bar it makes every other bar unreadable |
| **Frequency vs intensity** bubble chart | Active days a month against interactions per active day, per department — separates frequent-but-shallow from deep-but-occasional, which an adoption percentage cannot |

All dependency-free SVG, consistent with the existing `charts/` components. No new packages.

Habit buckets are normalised to a 28-day month and rounded to whole days, so "11+ active days" does not silently mean a near-daily user at 28 days and a fortnightly one at 180. `HabitBuckets_MeanTheSameThingWhicheverPeriodIsSelected` and `BucketBoundaries_MatchTheirPrintedCaptions` pin that.

---

## Review fixes folded in

Two independent reviews (Opus 4.8 and GPT-5.6) were run over the diff; everything they raised is fixed in this branch.

| Finding | Fix |
|---|---|
| **Treemap laid rows along the longer side** of the remaining space. Fills the area exactly, but produces slivers — the opposite of what squarification is for | Lay along the shorter side. Measured on representative data in the real viewBox: worst aspect ratio **14.49:1 → 2.11:1**, mean 3.75 → 1.41, area fill 1.0000 in both |
| **Bubble radius added a constant after the sqrt**, breaking the area proportionality the caption claimed — 100 vs 25 seats rendered as 2.6x rather than 4x | Pure sqrt scaling with a 3px floor only |
| **Largest bubble sat exactly on the axis maximum** and was clipped by the viewBox along with its label | 20% axis headroom + more top padding |
| **Displayed engagement formula omitted the `/ weightSum` division** the backend performs, so tuned weights would show components summing to ≠100% | Weights quoted as normalised shares; formula shows the divisor |
| **Intensity bubbles coloured by the whole-department mean** while both axes were active-only, contradicting the caption | Renamed `ActiveUserAverageScore`, averaged over active users. The department table keeps its whole-population mean; `IntensityBubbleColour_AveragesTheSamePopulationAsItsAxes` pins both |
| **`ActionDescription` called with default options**, not the ones the analysis ran with | Passes `_options` |
| **Hard-coded `28`, `"28 days"` and `75/50/25`** in explanations and scatter colours | Read from `summary.options`; `scoreColour` takes optional thresholds |
| **Donut centre showed the licence count** while its slices summed to the scored rows (differ if `MaxLicensedUsersScored` is hit) | Centre is the sum of the slices; the action-plan tip names its denominator |

## Testing

- `Tests.UnitTests` — **73 passing**, 9 new. New coverage: options serialisation casing, bucket boundaries against their printed captions, action-plan totals summing to the scored population, tag/prose agreement across every band, and the intensity averaging population.
- `npx tsc --noEmit` clean; the SPA builds as part of the `Web` project build.
- No schema migrations, no config-schema change, no new NuGet or npm packages.

## Reviewer hotspots

- `TreemapChart.tsx` `squarify()` — the row-orientation fix is the one place a subtle regression would be invisible in a screenshot.
- `CopilotAdoptionScoring.RecommendedActionCode` vs `RecommendedAction` — two switch statements that must stay in lockstep; the test covers every band, but a new band would need both updated.
- `BuildIntensity` uses `sum(interactions)/sum(activeDays)` (a weighted mean) rather than a mean of per-user ratios. That is intentional — it answers "on a day this department uses Copilot, how much do they do?" — but it is worth a second opinion.

## Deferred

Recorded in the wiki page's "Known gaps" section rather than done here: agent-level adoption, licence cost modelling, the scheduled adoption e-mail, unassigned purchased seats, and onboarding-cohort comparison.

---

# Changes since the original review

The five sections above describe the first commit. The branch has grown considerably since; this
section covers the remaining ten commits, newest work last.

## 6. Capability parity with `microsoft/AI-in-One-Dashboard` (`d47718e`)

A capability matrix against AI-in-One showed we covered roughly the licensed-seat third of it. Agents
(5 of their pages) and the unlicensed population (3 pages) were absent entirely.

Added: an **Agents** tab (inventory with Keep / Review / Retire / New verdicts), an **Unlicensed usage**
tab, usage concentration, a combined licensed-vs-unlicensed leaderboard, a three-population weekly
trend, and accessed resource types. The matrix is recorded in the wiki (§5a) so a future change cannot
silently regress a capability.

## 7. Executive visuals and the Excel export (`e0a028d`)

Funnel, gauges, radar, stacked area and quadrant visuals, plus a **12-sheet `.xlsx` export with live
Excel charts** — real charts, not pictures, so they can be re-plotted and dropped into a board pack.

The writer is **dependency-free OpenXML** written for this repo rather than a NuGet package: EPPlus 5+
is commercially licensed, and this solution has strict binding-redirect rules that make adding a
package to `Common.Entities` expensive. Lives in `Common/Entities/Xlsx/`.

## 8. Bug: `Invalid column name 'department'` (`096ac06`)

Two hand-written queries referenced `u.department`. `dbo.users` has no such column — it has
`department_id`, an FK to `dbo.user_departments.name`. Four occurrences across two queries, fixed with
the `LEFT JOIN dbo.user_departments` pattern already used correctly elsewhere in the same file.

The regression test was **proved to catch it**: reverting the fix made
`AgentAndUnlicensedQueries_ResolveDepartmentsThroughTheLookupTable` fail with the exact
`SqlException: Invalid column name 'department'`, four times over.

## 9. Correctness: capped denominators (`e1fc4d7`)

Every rate divided a **capped** count by the **uncapped** seat count. Above `MaxLicensedUsersScored`
that is not approximate, it is wrong: a 200k-seat tenant scored 50k deep could never report adoption
above 25%, and the funnel opened with a 75% drop that was pure measurement artefact.

`ScoredUsers` is now the denominator, `LicensedUsers` keeps its meaning as the true seat count, and
wherever the two differ the page, the warnings and the workbook all say so. Four chart rendering bugs
fixed in the same commit (funnel label invisible on narrow stages, gauge band misalignment, stretched
treemap glyphs from `preserveAspectRatio="none"`, squashed scatter labels).

## 10. Agent KPI window mismatch and band denominators (`7d35851`)

Found independently by two of three parallel model critiques: agent **interactions-per-user** divided a
120-day numerator by a 28-day denominator, inflating the headline figure roughly 4x. `AgentUsageSql`
now returns `WindowInteractions` alongside the history total so both sides of the ratio use one window.

Separately, the workbook's band percentages divided by `LicensedUsers` while the on-screen donut
divided by `ScoredUsers` — the export disagreed with the page it claimed to be a snapshot of. Both use
the analysed denominator now.

## 11. Overview restructure, named actions, drill-through (`e11c186`)

From the product critique: the Overview was ~18 cards in an order that buried "adoption by department"
near the bottom despite the spec naming the budget owner as the primary reader. Reordered into named
sections, "Coach"/"Grow" disambiguated, and every enablement-plan row is now a **drill-through** that
opens the licensed-user list filtered to exactly the people that number counted — filtered on the
action code rather than the band, because an action such as "Add a second app" spans two bands.

---

# Latest work

## 12. Models file split, and the reporting window pinned (`1818dbb`)

**Reorganisation — pure moves, no code edits.** `CopilotAdoptionModels.cs` held 29 unrelated types in
1,486 lines. Split along its existing `#region` seams into one file per concern; `CopilotAdoptionAnalysis`
lifted out of the 1,320-line service file so the feature's main result type is findable. The namespace
is unchanged, so no call site moved. A verification pass confirmed all 29 type declarations and every
code line carried over before anything was written.

The other four large files (`Sql`, `Scoring`, `Service`, `Workbook`) are each a **single cohesive class**
and were deliberately left alone — splitting those into partials would scatter one class across files
rather than help.

**Window fix — the one behaviour change.** "Last N days" started at midnight N days ago, which spans
**N+1** distinct calendar dates. A user active every single day therefore recorded 29 active days in a
"28-day" window, while the frequency component divides by a target derived from 28 — inflating the
score of precisely the most engaged users, and making the dates shown on screen a day wider than their
own label. The definition now lives in `CopilotAdoptionScoring.WindowStartUtc`, immediately beside
`TargetActiveDays`, so the numerator's window and the denominator's target cannot drift apart. The
agent inventory's reported `HistoryDays` had the same off-by-one and is fixed with it.

## 13. The engagement radar made readable (`cfd859b`)

Kept as a radar — it is the right chart for showing the *shape* of adoption, where two populations
average the same score with completely different profiles — but the three things that made it chaotic
are fixed:

1. **Two translucent filled polygons** produced a third, muddy colour exactly where the reader was
   comparing them. Only the first series fills now; the benchmark is a dashed outline, which also means
   the two are distinguishable without relying on colour.
2. **The rings were unlabelled**, so there was no anchor for whether a shape was big or small. They now
   carry their values, with a background-coloured halo so they stay legible over gridlines.
3. **The legend printed `45 / 30 / 22`** plus a footnote giving the axis order, leaving the reader to
   map three numbers onto three spokes. Replaced by a per-axis table carrying both series and a signed
   **Gap** column — which is the actual finding the card exists to show.

This also answers the area-exaggeration objection honestly rather than by deleting the chart: radar
area grows with the square of the values, so the outline overstates the difference. The shape is there
to be recognised, the table to be read, and no conclusion depends on estimating an area.

## 14. FakeDataGen: adoption personas (`467f08f`)

The combined profiling generator picked a **random user at a random timestamp** for every event. That
is correct for importer and volume testing, but it gives every licensed user near-identical usage: the
whole tenant lands in one band, and the funnel, the department treemap, the profile radar and every
"who should I target" panel correctly show nothing. The report looked broken when it was working.

New adoption-scenario mode (offered by *Generate combined profiling data*, default on) plants:

- every funnel stage, Champion through Never used;
- three distinctly different engagement **shapes** at similar scores — frequent-but-shallow (63),
  deep-but-narrow (58), broad-but-occasional (49) — which is exactly what the profile radar exists to
  tell apart;
- departments assigned Leading / Progressing / Lagging maturity, so the department charts show contrast;
- seats still held by disabled accounts;
- an unlicensed population with real usage, to rank as licence candidates;
- one agent per health verdict (Keep / Review / Retire / New).

Persona targets are expressed **only** in the three signals the analysis measures — distinct active
dates, interactions per active date, distinct `app_host` values — so the bands are produced by the real
scoring code rather than asserted by the generator. On every run the generator re-scores each persona
through `CopilotAdoptionScoring` and warns if any no longer lands in the band it claims, so a change to
a tuning default cannot silently turn the demo tenant into something that no longer demonstrates what
the wiki says it does.

Also extracts `CopilotChatFactory` so the scatter and the scenario build interactions through one path
instead of two copies that would drift, and drops a per-event `SELECT` that checked a freshly generated
GUID for collisions — one database round-trip per generated event.

All generated names are synthetic (`Contoso`).

---

## Testing (updated)

- **Copilot adoption suite: 109/109 passing**, including 2 new regression tests pinning the window to
  exactly N distinct dates.
- Full unit-test run: **945 passed**, 12 failed — all 12 are pre-existing environmental failures
  requiring credentials this machine does not have (Cognitive Services key, a Graph tenant, and a local
  `InstallerTestConfig.json`). None are in code touched by this branch.
- `Common.Entities`, `Tests.FakeDataGen` and `Web` all build; admin SPA `tsc --noEmit` clean.
- The models split is proved behaviour-neutral by the compiler plus the unchanged test suite.

## Reviewer hotspots (updated)

- `CopilotAdoptionScoring.WindowStartUtc` — the only behavioural change in commit `1818dbb`. Worth
  confirming you agree "last 28 days" should mean 28 dates including today.
- The models split is large but mechanical; reviewing `Entities.csproj` and confirming the build is
  probably more valuable than reading the moved files.
- `CopilotAdoptionPersonas` — the persona arithmetic is documented per persona and independently
  verified, but the *mix weights* per maturity tier are a judgement call.

## Deferred (updated)

Recorded in the wiki backlog rather than built here:

- **Pre-aggregated rollups.** One cold analysis is estimated at ~0.9bn chat-row visits at 200k users.
  This is the largest piece of engineering on the list and needs migrations plus the before/after
  benchmarks the schema-change policy requires.
- **Copilot Credits cost analytics.** `copilot_chats.copilot_credit_estimate_total` and its `_json`
  breakdown are already imported and unused. Deliberately **not** built yet: the wiki records rollups as
  the prerequisite for credit analytics being affordable at scale, so building it first would add more
  raw-fact scans to the hot path.
- Workbook streaming rather than buffering; caching compact summaries rather than the whole scored
  object graph.

---
---

# Performance due diligence

The repo's schema-change policy requires a measured before/after for anything performance-motivated.
This section is that evidence. The headline change here is **not** a schema change — it needs no
migration and no index — but the same measurement bar was applied, because the fix was chosen on
evidence rather than on intuition, and the first intuition was wrong.

## Test rig

| | |
|---|---|
| Schema | The **real** schema, built by applying all 51 EF migrations to an empty database. Not a hand-written subset — the point is to measure against the indexes a customer actually has. |
| Users | 200,000 |
| Copilot licences | 60,000, of which 34,000 have activity |
| Interactions | **12,000,000** `copilot_chats` + 12,500,000 `audit_events`, spread over 365 days, recent-weighted |
| Distribution | Skewed — median 245 interactions per active user, busiest 109k — so the per-user `GROUP BY` sees realistic cardinality (~34,000 groups) |
| Server | SQL Server 2025, 8 cores, MAXDOP 8 |
| Method | The real query text; `SET STATISTICS IO, TIME`; medians of 3 runs discarding the cold run; `OPTION (RECOMPILE)` (already in the shipped query) defeating plan caching; two selectivities |

No customer database was used, and no real data appears anywhere in this branch.

## What was actually wrong

The per-user aggregate asked for **three `COUNT(DISTINCT …)` in one `GROUP BY`**. SQL Server streams a
*single* distinct aggregate cheaply, but two or more force it to fan the input out through a spool and
process each distinct separately. That spool — not the fact scan — was essentially the whole query.

It also pushed the query past `CopilotAdoptionService.QueryTimeoutSecs` (90s), so on a large tenant this
was **not merely slow: the report degraded to a warning**.

### Isolation — one variable at a time

| Variant | Logical reads |
|---|---|
| Shipped | 37.2M |
| Narrow `app_host` from `nvarchar(MAX)` to `nvarchar(100)` | 36.9M — **~1%** |
| Remove *one* of the three distinct counts | 36.6M — barely moves |
| Remove **all three** | **92k** |
| A single distinct count only | 92k, **zero worktable** |

The intuitive suspect was wrong. `app_host` is an `nvarchar(MAX)` column being fed to
`COUNT(DISTINCT …)`, which looks exactly like the LOB-column anti-pattern the repo guidance warns
about — and narrowing it is worth about one percent. The last row is the real finding: **one** distinct
aggregate spools nothing; **two or more** spool.

## Result

| Window | | Logical reads | of which worktable | Elapsed |
|---|---|---|---|---|
| **28 days** (default) | before | 114,750,742 | 25,413,875 | **281 s** |
| | after | **772,272** | 232 | **73 s** |
| | | **−148×** | | **−3.8×** |
| **365 days** (widest) | before | 115,172,403 | 25,835,536 | **303 s** |
| | after | **818,851** | 792 | **135 s** |
| | | **−141×** | | **−2.2×** |

Both metrics improve at both selectivities, and the spool operator disappears. The default window now
sits comfortably inside the 90s timeout.

**Not claimed as solved:** the 365-day window is still 135s at this scale. It no longer fails outright,
but it is not comfortable, and it is the case the pre-aggregated rollups in the wiki backlog would
address. Those are re-scoped rather than dropped — still the right answer for the widest windows and
still the prerequisite for credit analytics, but no longer the *first* fix.

## Correctness was gated before performance

A 148× reduction is worthless if the numbers move. The old and new aggregates were run side by side
over the same window and compared row for row:

```
shipped_rows | rewrite_rows | differing_rows | verdict
       33999 |        33999 |              0 | IDENTICAL
```

That comparison exists specifically because the risk is NULL handling:
`COUNT(DISTINCT CASE WHEN … END)` ignores NULLs implicitly, whereas the rewrite has to exclude them
explicitly with `WHERE … IS NOT NULL`. Getting that wrong would silently inflate `AppsUsed` and
`AgentsUsed` for every user, which feeds the breadth component of the engagement score and therefore
the band a user lands in.

`LicensedUsersQuery_NeverAsksForSeveralDistinctCountsInOneGrouping` pins the shape in CI, because
re-introducing `COUNT(DISTINCT …)` here is an innocent-looking edit whose damage only shows up on a
tenant nobody tests against.

## A correction to my own method

The first run of this benchmark was invalid and is worth recording so the number is not quoted later.

The data generator picked each row's user through a `CROSS APPLY` whose predicate used `NEWID()` but
never referenced the outer row. With no correlation, SQL Server is entitled to evaluate that subquery
once rather than per row — and did, so all 500,000 rows in a batch received the same user. The result
was **25 users holding 500,000 interactions each** instead of 34,000 users with a realistic spread,
which makes a per-user `GROUP BY` benchmark meaningless.

It surfaced only because the correctness comparison returned 24 rows instead of tens of thousands. The
dataset was repaired by deriving the user from the row's own id (`CHECKSUM(au.id)`, a genuine per-row
expression), and **every number above is from the repaired dataset**. The same class of bug — a
non-deterministic expression that looks per-row but is not — also broke `CHOOSE()` in the generator
earlier and is now commented at both sites.

## The other three queries — measured, fixed, and a mixed result

`AgentUsageSql`, `UnlicensedUsageRowsSql` and `WeeklyAdoptionTrendSql` use the same
multiple-distinct shape. I expected a smaller blow-up because they group far more coarsely — per
agent, per week — but "should be smaller" is exactly the reasoning that was wrong about `app_host`,
so they were measured rather than assumed. **The guess was wrong for the second time.**

| Query | Distinct aggregates | Logical reads | of which worktable | Elapsed |
|---|---|---|---|---|
| `AgentUsageSql` | 4, per agent | 6,370,415 | 6,044,946 (95%) | **128 s** |
| `UnlicensedUsageRowsSql` | 3, per user | 8,404,506 | 8,124,681 (97%) | **131 s** |
| `WeeklyAdoptionTrendSql` | 4, per week | 24,468,573 | 24,163,354 (99%) | **315 s** |

Coarser grouping did not save them: the worktable is 95–99% of every one, and all three sat past the
90s timeout — so the agent inventory, the unlicensed population and the trend chart were each
degrading to a warning too.

`UnlicensedUsageRowsSql` is measured and verified with an **empty seat-licence list**, which makes
every active user count as unlicensed. That is deliberate — it is the worst case, and it is also the
real one the code comments call out: the "should we buy Copilot at all?" tenant has no Copilot SKUs,
and that is precisely when this query matters most.

### Two rewrites, chosen to suit the aggregate

- **`WeeklyAdoptionTrendSql`** collapses to one row per `(week, user)` first, carrying the per-user
  facts as flags. Every series there is a distinct *user* count under a different filter, and a user
  is licensed or not for the whole week — so `MAX()` over a flag gives the same answer and the roll-up
  becomes a trivial `SUM` with no distinct at all.
- **`AgentUsageSql`** and **`UnlicensedUsageRowsSql`** compute each distinct count from its own
  pre-projected set, as `LicensedUsersSql` now does.

### Result — and an honest regression

| Query | Logical reads | | Elapsed | |
|---|---|---|---|---|
| | before | after | before | after |
| `AgentUsageSql` | 6,370,415 | **2,735,140** ✅ | 128 s | **23 s** (−5.4×) |
| `UnlicensedUsageRowsSql` | 8,404,506 | **44,071,554** ⚠️ **+5.2×** | 131 s | **54 s** (−2.4×) |
| `WeeklyAdoptionTrendSql` | 24,468,573 | **58,061,724** ⚠️ **+2.4×** | 315 s | **38 s** (−8.4×) |

**This is not a clean win and is not being presented as one.** All three now finish well inside the
90s timeout, but two of them read substantially *more* pages than before. The spool is gone; what
replaces it is re-scanning the projected CTE once per distinct set, because SQL Server does not
materialise a CTE. Trading reads for time is normally the wrong way round, and this repo's own
guidance is right that a reads regression is the early warning that wall-clock will regress under
concurrency.

It is shipped anyway because the alternative is worse: the previous shape did not read fewer pages
*slowly*, it **exceeded the command timeout and returned no report at all**. A feature that works
while reading more pages beats a feature that fails. Reviewers who disagree with that trade-off
should say so — it is a judgement call, not a measurement.

The proper fix for the extra reads is to materialise the projected rows once rather than re-scanning
them, or to pre-aggregate at import time — which is exactly the **rollups** item in the wiki backlog.
This measurement is now the concrete argument for it, where before it was an estimate.

### Correctness, gated before performance

| Query | old rows | new rows | differing |
|---|---|---|---|
| `AgentUsageSql` | 200 | 200 | **0** |
| `UnlicensedUsageRowsSql` | 33,999 | 33,999 | **0** |
| `WeeklyAdoptionTrendSql` | 27 | 27 | **0** |

## Summary of the whole exercise

Four queries carried the same defect. All four exceeded the 90-second command timeout on a
200,000-user tenant, so the Copilot Adoption report was **not slow at scale — it was broken**, and
the failure mode was a warning rather than an error, which is why it had not been noticed.

| Query | Elapsed before | Elapsed after |
|---|---|---|
| `LicensedUsersSql` (28d) | 281 s | **73 s** |
| `LicensedUsersSql` (365d) | 303 s | **135 s** |
| `AgentUsageSql` | 128 s | **23 s** |
| `UnlicensedUsageRowsSql` | 131 s | **54 s** |
| `WeeklyAdoptionTrendSql` | 315 s | **38 s** |

No schema change, no migration, no new index — every one of these is a change to how the question
was asked. Two open items remain honestly on the table: the 365-day `LicensedUsersSql` window at
135 s, and the logical-read regression on two of the queries. Both point at the same next step.

---

# Telemetry

The feature had no telemetry beyond an exception handler on the workbook endpoint. That matters more
here than for most pages, because the failure mode is quiet: a query past the 90-second command
timeout **degrades to a warning on the page rather than an error**, so a tenant can sit with a
half-populated report indefinitely and never raise a ticket. Everything in the performance section
above was invisible in production, and would have stayed invisible.

Every completed analysis now emits one **`CopilotAdoptionAnalysis`** custom event:

| | |
|---|---|
| `customDimensions` | `WindowDays`, `WarningCount`, `Outcome` (`Complete`/`Degraded`), `TimedOut`, `SlowestStep` |
| `customMeasurements` | `TotalMs`, plus `<Step>Ms` per step — `LicensedUsersMs`, `WeeklyTrendMs`, `AgentEstateMs`, `LicenceOpportunitiesMs`, `UnlicensedPopulationMs`, `UsageByAppMs`, `ResourceTypesMs`, `ScoringMs` |

Four decisions worth a reviewer's attention:

- **Durations are measurements, not dimensions.** App Insights stores measurements as numbers, so
  `percentile(customMeasurements.TotalMs, 95)` works directly; the same value as a dimension is a
  string every query has to cast first.
- **One event per *analysis*, not per request.** Emitted inside the cache's `Lazy` factory. A cache
  hit does no work and must not look in telemetry as though it did, or the p95 becomes dominated by
  cached reads and the slow runs — the entire point of measuring — disappear into the average.
- **Timed in a `finally` block.** A step that threw, or that a timeout degraded into a warning, still
  records its duration. A query abandoned at 90 seconds is the most useful thing in this telemetry;
  recording only successful steps would hide exactly the tenants it exists to find.
- **Step names are constants** (`CopilotAdoptionSteps`) so an operator's saved query or alert survives
  a refactor.

Nothing in the event derives from tenant data — every value is a duration, a count, or a compile-time
constant. Telemetry is wrapped so a logging failure cannot change request behaviour, the same rule the
workbook endpoint already followed.

`AnalyticsLogger` gains a `TrackEvent` overload carrying metrics, following the shape of the existing
`TrackHealthCheck` / `TrackImporterHeartbeat` helpers; the existing two-argument overload is unchanged.

Wiki section 9 documents the event with ready-made Kusto for the three questions an operator actually
has — is any report degrading, how slow is it and which step owns it, is it approaching the timeout —
plus two suggested alert rules.

**Not verified:** the event shape is covered by unit tests and the build, but has not been observed
landing in a live Application Insights resource — that needs a deployed instance with a connection
string. Worth a sanity check on first deploy.

---

# Reviewer summary

| Area | Commits |
|---|---|
| Explainability, AI-in-One parity, Excel export, earlier fixes | `d44552d` … `e11c186` |
| Models file split (pure move) + reporting-window fix | `1818dbb` |
| Radar readability | `cfd859b` |
| Funnel: last step is not a loss, hover help | `fbae907` |
| Terminology: "seat" → "licence" in display text | `bd88a6f` |
| **SQL: remove multi-`COUNT(DISTINCT)`** | `a8da585`, `2e432ce` |
| App Insights telemetry | `be9f642` |

**Where to look hardest:** the two SQL commits. They change how four hot queries are written, and the
risk is NULL handling — `COUNT(DISTINCT CASE WHEN … END)` ignores NULLs implicitly while the rewrites
must exclude them explicitly. Each was gated by an old-vs-new row-for-row comparison at synthetic
scale (33,999 / 200 / 33,999 / 27 rows, zero differing), and `LicensedUsersQuery_NeverAsksForSeveral
DistinctCountsInOneGrouping` pins the shape in CI.

**Two things deliberately left open**, both pointing at the same next piece of work (pre-aggregated
rollups):

1. `LicensedUsersSql` at a 365-day window is still 135 s at 200k-user scale — no longer failing, but
   not comfortable.
2. Two of the rewrites trade **more logical reads** for much less elapsed time. That is normally the
   wrong way round; it is shipped because the alternative returned no report at all. A reviewer who
   disagrees with that trade should say so — it is a judgement call, not a measurement.


sambetts:
## Second pass: capability parity with AI-in-One Dashboard

Compared all 14 report pages of [microsoft/AI-in-One-Dashboard](https://github.com/microsoft/AI-in-One-Dashboard) one by one. The licensed-seat third was already covered; the agent estate and the unlicensed population were not covered at all. Both are now in, plus concentration and a combined leaderboard.

Full matrix in the wiki: [Copilot Adoption Tool - capability coverage](https://github.com/pnp/Microsoft365-Analytics-Insights/wiki/Copilot-Adoption-Tool#5a-capability-coverage-vs-the-ai-in-one-dashboard).

### New: Agents tab (5 of their pages; we had none)
Active/known/custom counts, most-used and most-versatile agent, inventory-health donut, interactions-per-agent treemap, agent usage by department, and a full inventory table with a **Keep / Review / Retire / New** verdict and a stated reason per agent.

`New` exempts anything first seen in the last 30 days from review entirely - a brand-new agent with two users has not failed, it has not started. The inventory spans the full 365-day history rather than the selected period, because an agent nobody has touched for six months is exactly what an inventory review is looking for and is invisible in a 28-day window.

### New: Unlicensed usage tab (3 of their pages; we had none)
These users previously existed only as licence candidates. Now reported in their own right - headline figures, their own habit strip using rules identical to the licensed one, usage by app, usage by department.

Uses a **separate query** from the candidate ranking, which is capped and score-ordered and therefore a biased sample that must never be used to describe a population's shape.

### New: usage concentration
Active licensed users ranked and cut into percentile cohorts with each cohort's share of activity. This is what an adoption percentage hides: *40% spread evenly* and *40% where a tenth do most of it* are the same number and completely different situations - the second collapses when those people change team.

### New: combined licensed-vs-unlicensed leaderboard
Per department, side by side, with conditional shading on both intensity columns. A department with idle seats **and** heavy unlicensed use is a seat-allocation problem, not an adoption problem - invisible in any single-population view.

### Trend and grounding
The weekly trend query now returns all three populations plus licensed/unlicensed volume in **one pass** over the same rows. Headcounts and volumes are split into two charts because plotting them on one axis flattens the headcount line onto zero. Added "What Copilot is working on" from accessed resource types - the clearest evidence Copilot is grounding on tenant content rather than answering generic questions.

### Deliberately not copied
- **Their AI-generated narrative page.** Needs an LLM call over tenant data and produces prose that reads authoritatively while being unverifiable - the exact opposite of this tool's first design principle.
- **A per-user leaderboard naming individual heavy users.** Ranking named individuals against each other is a stated non-goal. Every per-user list here exists to decide whether a *seat* is well placed, which is why the default sort is least-engaged-first.

### Where this goes further than theirs
Graded engagement score and bands (theirs is raw active-day buckets only), the adoption funnel, per-user recommended actions and the aggregated enablement plan, per-user licence justifications, the SKU classification table, SQL popovers, the methodology tab, and CSV export of both lists.

### Testing
**86 passing, 13 new** - agent verdict boundaries and the New exemption, agent users not double-counted across agents, most-popular vs most-versatile answering different questions, concentration accounting for every active user exactly once across nine population sizes, concentration ignoring idle seats, unlicensed habit bucketing matching licensed, and the combined view surfacing idle-seats-plus-heavy-unlicensed-use.

No schema migrations, no config-schema change, no new packages.

### Reviewer hotspots
- `WeeklyAdoptionTrendSql` changed from an INNER to a LEFT join on the seat CTE to pick up unlicensed users in the same pass. Same scan, broader result - worth a second opinion on the plan.
- `AgentUsageSql` groups over the history window rather than the reporting period; on a large tenant that is the widest scan this tool does.
- `BuildCombinedSegments` requires `MinSeatsPerSegment` of *either* population, not both - a department with no seats but many unlicensed users is a legitimate row.

sambetts:
## Third pass: executive-grade visuals + full Excel export

### Visuals — the page was almost entirely bars

All dependency-free SVG, so each is a deliberate choice rather than a drag-and-drop from a chart library.

| New | Why this shape |
|---|---|
| **Tapering funnel** (replaces bars) | Each stage is a trapezoid whose bottom edge matches the next stage's top edge, so the outline is continuous. The narrowing *is* the message — five bars of decreasing length made the reader compare lengths to reach the same conclusion |
| **Arc gauges** for adoption / habit / Cowork rate | A bare percentage gives no sense of whether it is good. Banded, not a smooth gradient — a continuous ramp implies the difference between 41% and 43% means something |
| **Radar** of the frequency/depth/breadth profile | Typical active user vs the tenant's *own* Champions. Two tenants can average the same score with opposite shapes and need opposite programmes; the gap between the outlines says which behaviour to target |
| **Stacked area** for population composition | Lines compare series; a stacked area shows the total and its make-up at once. The line chart is kept alongside rather than replaced, because only the bottom band can be read precisely |
| **Median quadrant lines** on the intensity scatter | Thresholds are the tenant's own medians, not fixed values — "busier than half your own departments" is defensible on any tenant. Four quadrants named |
| Doughnut beside the habit tiles, gradient rounded bars | Polish; the tiles read a bucket precisely, the ring shows the balance |

Two of these — the **radar** and the **gauges** — have no AI-in-One equivalent at all.

### Excel export — the point-in-time snapshot

New **Excel report** button. Twelve sheets, **native live Excel charts bound to cell ranges**, not pictures.

The purpose is comparing before and after an enablement programme. A screenshot cannot be compared with another screenshot six months later — the numbers cannot be subtracted and nobody can tell what period or thresholds either was run with. The **Report** sheet records both, and the methodology sheet states explicitly that two files are only comparable if the thresholds match, because the tuning is adjustable and *"adoption went up"* must not turn out to mean *"the bar moved"*.

**Zero new dependencies.** `Common.Entities.Xlsx` is a small hand-rolled OpenXML writer over `System.IO.Compression`. EPPlus 5+ is commercially licensed, and any third-party library would have meant binding-redirect changes across several projects per the repo's NuGet rules.

Because it is hand-rolled, `CopilotAdoptionWorkbookTests` generates a **real workbook on every build** and asserts: valid OPC package, every part well-formed XML, no dangling relationship, charts reference cell ranges rather than images, Greek + ampersands survive, and numbers written invariantly **under a `de-DE` culture**. A malformed part reaches a customer as "Excel found unreadable content" with no clue as to the cause.

### Review fixes folded in

Reviewed by a separate agent; everything it raised is fixed:

| Finding | Severity | Fix |
|---|---|---|
| `workbook.xml` declared no `<bookViews>` while every sheet emitted `workbookViewId="0"` — that reference dangled in **every file produced** | High | Declared, and asserted by a test |
| Sheet-name sanitising trimmed apostrophes *before* truncating to 31 chars, so a cut could expose a mid-name apostrophe as the last character (Excel rejects it) | Medium | Trim runs again after the cut; test added |
| Licensed-user sheet capped at 20,000 rows **silently** while the analysis allows 50,000 | Medium | Says so in the sheet title and a note, with both counts |
| Departments sheet returned early unless department/app data existed, but also renders country and resource-type sections — a tenant with only country data lost it entirely | Medium | All six sections in the guard |
| Funnel's final stage used a fixed minimum bottom width, so a zero/small Champions stage drew a trapezoid that **widened** at the bottom | Medium | Clamped so it can never exceed the top edge |
| Export endpoint let a hand-rolled-XML failure surface an exception and stack trace | Medium | Catches, logs via `AnalyticsLogger`, returns a plain message |
| Agent inventory table is 11 columns wide but its chart was anchored at `K3` | — | Found and fixed before review; moved clear |

Reviewer confirmed clean: numeric XML output is invariant throughout, user text is escaped, column-name/range arithmetic beyond Z is correct, stream lifetime is correct, and the gauge/radar/stacked-area maths, React hooks, routing and dependency policy showed no issues.

### Testing
**98 passing, 12 new.** No schema migrations, no config-schema change, no new NuGet or npm packages.

### Reviewer hotspots
- `XlsxChart.cs` builds chart XML by hand. The tests prove the package is structurally valid but cannot prove Excel renders every chart type attractively — worth opening one generated file if you have Excel to hand (this machine had neither Excel nor LibreOffice).
- Chart anchor cells in `CopilotAdoptionWorkbook` are still literals (`F3`, `K22`, `N22`). I checked each against its sheet's column count, but a future section added above a chart could push data under it.

Issue #329, opened 2026-08-26
Installer: Copilot AI interaction history import has no Test Configuration check, so a missing AiEnterpriseInteraction.Read.All grant is only discovered after install


sambetts:
## Summary

**Copilot AI interaction history (tenant-wide unless scoped)** — `ImportTaskSettings.CopilotInteractionHistory` — has **no Test Configuration check at all**. An admin can tick it, run the tests green, complete the install, and only find out hours later that nothing was imported.

That is a bad gap for this toggle in particular, because it is the **only** import whose permission the installer does not grant. `AiEnterpriseInteraction.Read.All` needs separate, explicit admin consent, so it is simultaneously the most likely toggle to be misconfigured and the only one Test Configuration says nothing about.

The failure is also **silent by design**. `CopilotInteractionHistoryImporter.ImportAsync()` checks the permission and, if absent, logs a warning and returns:

```csharp
if (!await _sourceLoader.HasInteractionReadAccessAsync())
{
    _logger.LogWarning(
        "Skipping the Copilot interaction history import: the runtime identity does not hold the " +
        "AiEnterpriseInteraction.Read.All application permission. ...");
    return null;
}
```

No error, no failed import, no health alert — just an empty table and a warning buried in web-job logs.

## What Test Configuration covers today

`SolutionInstallVerifier.VerifyRuntimeAccountAllAPIs` / `VerifyTeamsAndUserActivityImport`:

| Toggle | Verified? |
|---|---|
| `ActivityLog` | yes → `VerifyActivityAPIImport` |
| `GraphUsageReports` | yes → `VerifyUserActivityImport` |
| `GraphCopilotUsageReports` | yes (shares the `Reports.Read.All` check) |
| `GraphTeams` | yes → `VerifyTeamsImport` |
| **`CopilotInteractionHistory`** | **no — nothing** |
| **`Copilot`** | **no — see gap B below** |

## Gap A: `CopilotInteractionHistory` is never verified

Nothing in `SolutionInstallVerifier` references `CopilotInteractionHistory`. There is not even a "skipping ... not being targeted" line, so the omission is invisible in the test output — the admin has no signal that this workload went unchecked.

**The check already exists and is cheap.** `GraphAiInteractionSourceLoader.HasInteractionReadAccessAsync()` reads the `roles` claim off the app token via `GraphTokenPermissions.Extract`:

```csharp
var token = await _appIdentity.GetAccessToken();
[...]

PR #354, opened 2026-08-28
Installer: verify the Copilot import permissions Test Configuration silently skipped

sambetts:
Addresses #329 by making Test Configuration explicitly verify the Copilot AI interaction history permission and by adding a structural guard so future `[ImportProp]` toggles cannot ship without a verification decision.

### Copilot AI interaction history permission check

`SolutionInstallVerifier` now verifies `ImportTaskSettings.CopilotInteractionHistory` instead of silently omitting it. The check uses `GraphAiInteractionSourceLoader.GetInteractionReadAccessAsync()` against the runtime Graph app identity and reports a tri-state result:

| Result | Test Configuration output | Meaning |
|---|---|---|
| `Granted` | INFO | the issued Graph app token contains `AiEnterpriseInteraction.Read.All` |
| `NotGranted` | ERROR | the token was readable and lacks the role, so the import will skip at runtime |
| `Unknown` / `NoIdentityToInspect` | WARNING | the check could not prove the grant either way; it is not reported as a pass |

Important scope: this is an issued-token role check, not an app-registration manifest read. It catches the "permission configured on the app registration but not actually admin-granted for the tenant" case, because Entra will not emit `AiEnterpriseInteraction.Read.All` in the app-only Graph token until the tenant grant exists. It does **not** make a live `getAllEnterpriseInteractions` data-plane call, so it does not validate per-user service-plan eligibility.

When `CopilotInteractionHistory` is off, Test Configuration logs `Skipping verifying Copilot AI interaction history import as not being targeted`; it is a SKIP, not a PASS.

### Import-toggle coverage guard

`SolutionInstallVerifier.ImportToggleCoverages` declares every `[ImportProp]` toggle as either verified or explicitly unverifiable, and `ReportUnverifiedEnabledImports()` warns admins about enabled imports that Test Configuration cannot prove. `ImportToggleVerificationCoverageTests` asserts the table matches the actual `[ImportProp]` set, has no duplicates, and cannot contain contradictory entries.

### Deliberate deconfliction with #446 / #355

`UsesActivityApi` and the verifier's Activity API condition are intentionally left to PR #446 (#355) to avoid a conflicting duplicate definition. This PR no longer defines `ImportTaskSettings.UsesActivityApi`, no longer changes the web-job Activity API gate, and no longer changes Test Configuration's existing `ActivityLog`-only Activity API condition.

## Schema / config impact

- No EF migrations and no database schema change.
- No `CONFIG_VERSION` bump: no persisted property was added, removed, or renamed. The remaining `ImportTaskSettings` changes are helper methods used by the coverage test/reporting and do not alter the saved settings string or JSON schema.

## Validation

- Rebased onto current `origin/dev`; the deconfliction reverted the shared Activity API condition to `origin/dev` ownership.
- Local build: `MSBuild "O365 Advanced Analytics Engine.sln" /p:Configuration=Debug /m:2` passed.
- Local targeted tests under the shared vstest lock: `ImportToggleVerificationCoverageTests` passed (21/21).
- CI is the authoritative full-suite evidence for this batch: `test_dotnet (Release)` passed in 13m09s after the deconfliction push; `build_dotnet (Release, Web, ...)` passed in 2m45s.
- Multi-model code review: two earlier rounds across `claude-opus-4.8`, `gpt-5.6-sol`, and `grok-4.6`, plus a deconfliction re-review on the same three models; final deconfliction round was clean.

## Reviewer hotspots

- Confirm `HasInteractionReadAccessAsync` compatibility: null identity still fails open for the importer; unreadable/malformed tokens fail closed as before.
- Confirm the verifier reports `Unknown`/`NoIdentityToInspect` as warnings rather than false passes or false missing-grant errors.
- Confirm `UsesActivityApi` remains solely in PR #446 / #355.

Addresses #329

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,4 @@
-* @sambetts @jesusfer
+* @sambetts
 /.github/ @sambetts @jesusfer
 /reports/ @jesusfer
+/src/AnalyticsEngine/App.ControlPanel.Engine/SqlExtentions @jesusfer
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/BaseInstallProcessClasses.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/BaseInstallProcessClasses.cs
@@ -1,5 +1,6 @@
 ﻿using App.ControlPanel.Engine.Entities;
 using App.ControlPanel.Engine.Models;
+using CloudInstallEngine.Azure;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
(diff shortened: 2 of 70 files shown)


sambetts:
## Merge-order note (PR #354 + PR #446)

These two PRs both edit `VerifyRuntimeAccountAllAPIs` in `App.ControlPanel.Engine/SolutionInstallVerifier.cs`. A dry-run `git merge-tree` of the two branches reports **one content conflict, in that file only**:

```
CONFLICT (content): Merge conflict in src/AnalyticsEngine/App.ControlPanel.Engine/SolutionInstallVerifier.cs
Auto-merging src/AnalyticsEngine/Common/Entities/ImportTaskSettings.cs   <- clean
```

`ImportTaskSettings.cs` auto-merges cleanly, so the previously dangerous collision (both PRs defining `UsesActivityApi` with opposite semantics) is resolved: **#446 alone defines it**, and #354 no longer does.

**Recommended order: merge #446 first, then rebase #354.**

**Required resolution — keep BOTH sides:**

1. Keep #446's gate: `if (Config.SolutionConfig.ImportTaskSettings.UsesActivityApi)` (**not** `.ActivityLog`). #354 deliberately reverted this line to `ActivityLog` to avoid owning the member, so on rebase #446's version must win.
2. Keep #354's added `VerifyCopilotInteractionHistoryImport` call and its method.

**Do not** resolve by taking #354's side of the gate. That would silently revert #355 - a Power-Platform-only tenant would stop running the Activity API import - and CI would still pass, because no remaining test asserts the inclusion at that call site.

PR #358, opened 2026-08-28
Profiling: stop usp_UpsertCopilot dropping the last day of every week

### LINKED ISSUE #300: Review the profiling weekly-boundary change to usp_UpsertCopilot (reverted from the pending release)
## What was changed

`0d7b152` (a test-data-generator commit) altered the date predicate in
`profiling.usp_UpsertCopilot` in two places, from an inclusive `BETWEEN` to a half-open window:

```sql
-- before
WHERE @StartDate <= au.time_stamp AND au.time_stamp <= @EndDate

-- after
WHERE au.time_stamp >= @StartDate
  AND au.time_stamp < DATEADD(DAY, 1, @EndDate)
```

The commit message describes it as "include the full Sunday boundary in weekly aggregation".

It has been **reverted out of the pending release** so it can be assessed on its own terms.

## Why it was reverted rather than kept

The change is plausibly correct. If `@EndDate` is a date at midnight, `au.time_stamp <= @EndDate`
keeps only the single instant `00:00:00` of the end day, so the last day of the window is
effectively dropped — a real under-count.

But:

1. **It silently changes every weekly Copilot metric the proc produces.** The direction and size of
   the change depend on what callers pass as `@EndDate`, which is not obvious from the proc alone.
2. **If any caller already passes an exclusive end** (the start of the next period), the new form
   double-counts a day instead of fixing an under-count.
3. It arrived alongside a test-data generator, with no before/after figures, and profiling SQL is the
   area of this codebase that most needs deliberate handling — it feeds the weekly aggregates
   customers report on, and a silent shift shows up as unexplained movement in historical charts.

## What the review needs to establish

- [ ] What every caller of `usp_UpsertCopilot` passes as `@StartDate` / `@EndDate` — in particular
      `WebJob.Office365ActivityImporter/AutomationPS/ProfilingJobs/Weekly.ps1`. Is `@EndDate`
      inclusive (the Sunday) or exclusive (the following Monday)?
- [ ] Whether the same boundary bug exists in the **other** profiling upsert procs. If it is a real
      bug it is unlikely to be Copilot-only, and fixing one workload alone would make workloads
      disagree with each other.
- [ ] Before/after weekly counts on a populated database, so the size of the correction is known.
- [ ] Whether previously-compiled weekly rows need recompiling. If the boundary was wrong, history is
      wrong too, and a forward-only fix leaves a step change in the middle of every trend chart.

## Suggested outcome

If the fix is right, land it as a **standalone profiling change** with the caller audit, the same fix
applied consistently across workloads, before/after numbers, and a decision recorded about
recompiling history.

`Tests.FakeDataGen` can now generate Copilot and O365 profiling source data, so this is measurable
without a customer database.

sambetts:
Resolves the review #300 asked for, and lands the fix.

## The bug is real, and it is Copilot-only

`profiling.usp_UpsertCopilot` filtered with:

```sql
WHERE @StartDate <= au.time_stamp AND au.time_stamp <= @EndDate
```

`dbo.audit_events.time_stamp` is a **`datetime`** carrying a real time of day. `@EndDate` is a **`DATE`**. SQL Server widens `@EndDate` to `Sunday 00:00:00.000`, so `<= @EndDate` keeps only events at exactly that instant and **silently discards the rest of Sunday** — from every weekly Copilot metric the proc produces.

### 1. What does every caller pass as `@StartDate` / `@EndDate`?

**There is exactly one caller**, and `@EndDate` is **inclusive**:

```sql
-- usp_CompileActivityWeek
DECLARE @Sunday DATE = DATEADD(DAY, 6, @Monday);
...
EXECUTE profiling.usp_UpsertCopilot @Monday, @Sunday;
```

`Weekly.ps1` reaches this only through `usp_CompileWeekly @WeeksToKeep` and passes no dates of its own.

This settles the issue's main worry — *"if any caller already passes an exclusive end, the new form double-counts a day"*. No caller does, so the half-open window is a straight correction with no double-count. There is now a test asserting exactly that.

### 2. Does the same bug exist in the other profiling upsert procs?

**No — and the reason matters, because the surface reading says otherwise.**

Every `usp_Upsert*` proc uses the same `"date" <= @EndDate` shape, and I confirmed against a real database that **all** these columns are declared `datetime`, not `date`:

| Table | Column | Type |
|---|---|---|
| `audit_events` | `time_stamp` | `datetime` |
| `teams_user_activity_log` | `date` | `datetime` |
| `onedrive_user_activity_log` | `date` | `datetime` |
| `sharepoint_user_activity_log` | `date` | `datetime` |
| `outlook_user_activity_log` | `date` | `datetime` |

So the type alone doesn't explain it. What does is the **values**: the usage-report tables hold one row per user per *report day*, always at midnight, so `<= @EndDate` is exact for them. `audit_events` holds real event times — measured on a local synthetic set, **100% of 10,000,000 rows have a non-midnight time**.

`usp_UpsertCopilot` is the only proc aggregating raw audit events, and therefore the only one affected. Changing the others would be churn, not a fix — and would risk exactly the "silent shift" the issue warns about.

### 3. Before/after numbers

Measured on a local synthetic dataset (12M Copilot chats joined to `audit_events`), taking one Mon–Sun week and running both predicates over the same rows:

| Metric | Current (`<= @Sunday`) | Fixed (`< @Sunday + 1`) | Change |
|---|---|---|---|
| Copilot chats | 158,008 | 184,364 | **+16.7%** (old form lost 14.3% ≈ exactly 1/7) |
| Distinct active users | 32,361 | 32,950 | **+1.8%** |

Volume metrics lose a whole day. The distinct-user count loses less, because most Sunday-active users were also active earlier in the week — **but a user who only used Copilot on a Sunday vanished from that week entirely**, which is the more damaging error for an adoption metric.

*(Synthetic benchmark data — no customer figures.)*

### 4. Does previously-compiled history need recompiling?

**Yes.** There is no update path — a week is compiled once and never revisited — so existing weekly rows stay wrong and the fix would otherwise show up as a step change in the middle of every trend chart.

`PROFILING.md` now documents the procedure, including two things that are easy to get wrong:

- `@All = 1` only moves the **start** date back. The row-existence guard still skips weeks that already have rows, so the existing rows must be **deleted first** or the recompile is a no-op.
- Recompilation is bounded by raw-data retention: weeks whose `audit_events` / `*_user_activity_log` rows have aged out cannot be corrected.

Not automated — deleting and rebuilding a customer's weekly aggregates is an operator decision for a maintenance window, not something an upgrade should do silently.

## The fix

Both predicates in `usp_UpsertCopilot` (the `hosts_pivoted` PIVOT source and the `events` CTE) become half-open:

```sql
WHERE au.time_stamp >= @StartDate AND au.time_stamp < DATEADD(DAY, 1, @EndDate)
```

Both had to change together — otherwise the per-host counts and the chat/file/meeting counts would cover different windows for the same week. The reasoning is left in the SQL as a comment, since the next person to read `<= @EndDate` will assume it is fine.

## Tests

`UspCompileActivityWeek_CopilotSundayActivity_IsIncluded` was already in the suite as an `[Ignore]`d reproduction pointing at this issue ("*Un-ignore this as part of #300; it is the assertion that change needs to satisfy*"). It is now **enabled and passing**. Two boundary guards added alongside it:

- `..._CopilotActivityAfterTheWeek_IsExcluded` — activity at the following Monday `00:00` and later belongs to the **next** week. This is the over-count guard.
- `..._CopilotWholeWeekBoundaries_AreCounted` — seeds five interactions (first instant of the week, mid-week, last minute of Sunday, one second before the week, and the next Monday) and asserts exactly three are counted.

**Verified in both directions.** With the old predicate restored:

```
Failed UspCompileActivityWeek_CopilotSundayActivity_IsIncluded
  Assert.AreEqual failed. Expected:<1>. Actual:<0>.
Failed UspCompileActivityWeek_CopilotWholeWeekBoundaries_AreCounted
  Assert.AreEqual failed. Expected:<3>. Actual:<2>.
Passed UspCompileActivityWeek_CopilotActivityAfterTheWeek_IsExcluded
```

The over-count guard passing in *both* configurations is the point — it proves the fix corrects an under-count without introducing an over-count. With the fix: **19/19 profiling tests pass** against a real LocalDB running the actual stored procedure.

## Migration / risk

- **No EF migration.** The profiling schema is an idempotent `DROP`/`CREATE` SQL extension that the installer re-applies on every database upgrade, so the corrected proc ships with the normal upgrade.
- **No config change.**
- **Behavioural change is intended and now measured**: weeks compiled *after* the upgrade include the full Sunday. Weeks compiled before it do not, until recompiled — hence the documented procedure.
- Also documents the general rule in `PROFILING.md` §6 so the next proc that filters a timestamped table uses the half-open form.

## Reviewer hotspots

- The claim that **no other proc is affected**. It rests on the usage-report tables only ever holding midnight values. If any loader could write a non-midnight `date`, that proc would need the same fix.
- Whether **history recompilation should be automated** rather than documented. I've deliberately left it manual.

Addresses #300

### DIFF
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/SqlExtentions/PROFILING.md
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/SqlExtentions/PROFILING.md
@@ -181,6 +181,45 @@ counts chats/files/meetings via `dbo.copilot_event_files` / `dbo.copilot_event_m
   **does not re-raise**. So a failed week is logged but the run still returns success. The
   Profiling tab / `TraceLogs` are the only place failures show up.
 - **Idempotent installer, idempotent compile** (via the row-existence guards). Re-running is safe.
+- **Date predicates: watch the column type.** All the `usp_Upsert*` procs take `@StartDate DATE,
+  @EndDate DATE` and are called with `@Monday, @Sunday` — an **inclusive** Sunday. Most of them
+  read a daily `*_user_activity_log` table whose `"date"` column, although declared `datetime`,
+  only ever holds midnight values, so `"date" <= @EndDate` is exact. **`usp_UpsertCopilot` is the
+  exception**: it aggregates raw `dbo.audit_events`, whose `time_stamp` carries a real time of day.
+  There, `time_stamp <= @EndDate` compares against the Sunday widened to `00:00:00.000` and keeps
+  only the first instant of the day — it must use a half-open
+  `time_stamp < DATEADD(DAY, 1, @EndDate)` window. This was a live bug (issue #300): every weekly
+  Copilot metric under-counted by a full day (~1/7). If you add another proc that filters a
+  timestamped table, use the half-open form.
+
+---
+
+## 6a. Recompiling history after an aggregation fix
+
+There is no update path — a week is compiled once and never revisited — so a change to *how* a week
+is aggregated does **not** correct weeks already written. Until history is recompiled, a fix shows up
+as a step change in the middle of every trend chart.
+
+To recompile, delete the affected weeks and let the next run refill them, or force a full recompile:
+
+```sql
+-- Recompile every retained week from scratch (@All = 1 starts from the retention date).
+EXEC profiling.usp_CompileWeekly @WeeksToKeep = 53, @All = 1;
+```
+
+`@All = 1` only moves the *start* date back; the row-existence guard still skips weeks that already
+have rows, so the existing rows must be removed first for the recompile to take effect:
+
+```sql
+DELETE FROM profiling.ActivitiesWeekly       WHERE MetricDate >= @FromMonday;
+DELETE FROM profiling.ActivitiesWeeklyColumns WHERE "date"    >= @FromMonday;
+DELETE FROM profiling.UsageWeekly            WHERE "date"     >= @FromMonday;
+```
+
+Recompilation is only as good as the retained raw data: `usp_CompileWeekly` reads
+`dbo.audit_events` and the `*_user_activity_log` tables, so weeks whose raw rows have already been
+aged out cannot be corrected. Recompile in a maintenance window — it re-aggregates every retained
+week in one pass.
 
 ---
 
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/SqlExtentions/Profiling-03-CreateSchema.sql
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/SqlExtentions/Profiling-03-CreateSchema.sql
@@ -1905,7 +1905,15 @@ BEGIN
           event_id
         FROM dbo.copilot_chats AS c
           JOIN dbo.audit_events AS au ON c.event_id = au.id
-        WHERE @StartDate <= au.time_stamp AND au.time_stamp <= @EndDate
+        -- Half-open window, NOT "<= @EndDate". au.time_stamp is a datetime carrying a real time of
+        -- day, while @EndDate is a DATE - so SQL Server widens @EndDate to midnight and "<= @EndDate"
+        -- keeps only events at exactly 00:00:00.000 on the last day, silently dropping the whole of
+        -- it. The only caller (usp_CompileActivityWeek) passes @Monday, @Sunday with Sunday
+        -- INCLUSIVE, so every weekly Copilot metric was under-counting by a full day. Measured on
+        -- synthetic data: 158,008 chats vs 184,364 for one week (-14.3%, i.e. exactly 1/7).
+        -- The other usp_Upsert* procs are unaffected: they read the daily *_user_activity_log tables,
+        -- whose "date" values are always midnight, so the same comparison is exact for them.
+        WHERE au.time_stamp >= @StartDate AND au.time_stamp < DATEADD(DAY, 1, @EndDate)
       ) t
       PIVOT (
         COUNT(event_id)
@@ -1977,7 +1985,9 @@ BEGIN
         JOIN dbo.audit_events AS au ON c.event_id = au.id
         LEFT JOIN dbo.copilot_event_files AS f ON c.event_id = f.copilot_chat_id
         LEFT JOIN dbo.copilot_event_meetings AS m ON c.event_id = m.copilot_chat_id
-      WHERE @StartDate <= au.time_stamp AND au.time_stamp <= @EndDate
+      -- Same half-open window as hosts_pivoted above; the two must agree or the per-host counts and
+      -- the chat/file/meeting counts would cover different windows for the same week.
+      WHERE au.time_stamp >= @StartDate AND au.time_stamp < DATEADD(DAY, 1, @EndDate)
     ),
     event_counts AS (
       SELECT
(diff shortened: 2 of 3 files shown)


sambetts:
@jesusfer — tagging you as the original author of `profiling.usp_UpsertCopilot` (introduced in c022069, "PR: Add Copilot events data in the reports" #6). This PR changes the date predicate in that proc, so I'd value your read before it goes in.

## What this changes

Two `WHERE` clauses in `usp_UpsertCopilot`, from an inclusive comparison to a half-open window:

```sql
-- before
WHERE @StartDate <= au.time_stamp AND au.time_stamp <= @EndDate
-- after
WHERE au.time_stamp >= @StartDate AND au.time_stamp < DATEADD(DAY, 1, @EndDate)
```

## Why

`@EndDate` is a `DATE`, but `au.time_stamp` is a `datetime` carrying a real time of day. SQL Server widens `@EndDate` to midnight, so `au.time_stamp <= @EndDate` matches only rows at exactly `00:00:00.000` on the final day — the rest of that day is silently dropped.

The only caller is `usp_CompileUsageWeek`, which defines `@Sunday DATE = DATEADD(DAY, 6, @Monday)` — a Mon–Sun **inclusive** week. So every weekly Copilot metric was short by very nearly a full day out of seven. Measured on synthetic data: **158,008 chats vs 184,364 for one week (−14.3%, i.e. exactly 1/7)** — the arithmetic matches the theory, which is the main reason I believe the diagnosis.

This proc is the odd one out: the other `usp_Upsert*` procs read the daily `*_user_activity_log` tables whose `date` values are always midnight, so the inclusive comparison is exact for them and they need no change.

## Verification done before asking for your review

- **Boundary correctness.** The new window is `[Monday, Monday+7)`, which sits exactly adjacent to the following week's `[Monday+7, …)`. **No gap and no double-count** across consecutive weeks — worth stating explicitly, since that is the obvious risk when moving from inclusive to half-open.
- **Compatibility with the just-shipped 1810 release.** 1810 also modified `Profiling-03-CreateSchema.sql` (qualifying two `user_id` references that became ambiguous once `dbo.copilot_chats` gained its own denormalised `user_id`). This branch predates that, so I did not rely on the "no conflicts" badge — I materialised the merge result and checked it directly: the qualification survives (3 occurrences), this PR's half-open window applies (2 occurrences), and no instance of the old predicate remains. `ProfilingStoredProcedureTests.cs` was untouched by 1810, so the test changes here apply cleanly.
- **Currency.** The branch is 7 commits behind `dev` but merges cleanly and is semantically correct as-is, so it does not need a rebase.

## What I'd like your view on

1. Was the inclusive `<= @EndDate` deliberate — i.e. is there a reason the last day was meant to be excluded that I've missed?
2. Does the correction have any knock-on effect on the Analytics Report you're working on in #101? Weekly Copilot figures compiled by this proc will **increase by roughly 1/7** once this lands, and historical weeks stay wrong until recompiled, so any stored baselines or expected values in the report would shift.
3. Whether recompiling existing weeks should be part of this change or tracked separately.

jesusfer (approved):

Issue #167, opened 2026-06-26
Enforce uniqueness on dbo.urls.full_url (make IX_urls_full_url unique) to prevent duplicate URL lookups - future migration


sambetts:
## Summary
Make `dbo.urls.full_url` **unique** so duplicate URL lookups can never be created in the first place. Today `IX_urls_full_url` is a **non-unique** non-clustered index on `dbo.urls.full_url` (`nvarchar(850) NOT NULL`, latest migration `202606141000001_UrlFullUrlNvarchar`), so duplicate URLs accumulate in customer databases.

This is the source-level prevention that complements the defensive fix in #165: duplicate `urls.full_url` rows are what fan a single page-view out into multiple hit rows sharing one `page_request_id`, which used to break the unique `IX_PageRequestID` and abort a whole day's import. #165 makes the hits merge tolerate duplicates; **this** issue stops them being created.

## Proposed change (future migration)
1. **De-duplicate existing rows before creating the unique index.** Do **not** reuse `Resources/Delete Duplicate URLs.sql` unchanged: it is a cursor using `varchar(max)` variables (not Unicode-safe) and currently misses `hits_clicked_elements.url_id`. Implement a set-based, `nvarchar`-safe migration that maps every duplicate URL ID to a canonical survivor.
2. **Repoint every current reference before deleting duplicate URLs.** The current schema has FKs from:
   - `hits.url_id`
   - `copilot_event_files.url_id`
   - `file_metadata_property_values.url_id`
   - `hits_clicked_elements.url_id`
   - `page_comments.url_id`
   - `page_likes.url_id`

   Also repoint the legacy non-FK reference `event_meta_sharepoint.url_id` when that table/column exists. Discover/guard each table and column so the migration converges safely across database versions.
3. **Enforce uniqueness.** Drop `IX_urls_full_url` and recreate it as a UNIQUE non-clustered index (or add a UNIQUE constraint):
   `CREATE UNIQUE NONCLUSTERED INDEX IX_urls_full_url ON dbo.urls(full_url) WITH (IGNORE_DUP_KEY = ON, ONLINE = ON, ...)`.

## Considerations
- **Index-key limit:** keep `full_url` as `nvarchar(850)` — that is exactly the 1700-byte non-clustered index-key limit (2 bytes/char) and is Unicode-safe (Greek etc., see #122). A unique index fits but is at the limit (no room to widen).
[...]

Issue #165, opened 2026-06-26
AppInsights importer: duplicate IX_PageRequestID aborts whole day's hit import (JOIN fan-out on duplicate dimension rows)


sambetts:
## Summary
A customer's AppInsights importer fails an entire day's hit import with:

```
DataUtils.Sql.BatchSaveException: Couldn't merge batch insert using given SQL: Cannot insert duplicate key row in object 'dbo.hits' with unique index 'IX_PageRequestID'. The duplicate key value is (e72ee580-0344-4f74-8c98-c9e00d95089f).
...
Failed saving data for day 2026-05-18: ... Skipping this day and continuing.
```

## Root cause — JOIN fan-out in the hits merge
The final `INSERT INTO hits ... SELECT DISTINCT ...` in `WebJob.AppInsightsImporter.Engine/Resources/Migrate Hits Import into Hits.sql` joins the staging table to dimension tables (`urls`, `sessions`, `users`, `page_titles`, `webs`, `devices`, `browsers`, `operating_systems`, cities/countries/provinces).

None of those natural keys is unique:
- `urls.full_url` has only a **non-unique** `IX_urls_full_url` (`Common/Entities/Entities/Url.cs`).
- `sessions.ai_session_id` is **deliberately non-unique** (`Sql/ImportDbHacks.cs` — `EnsureSessionTableHasRightCollation` even downgrades a unique index to non-unique).

So when a dimension holds a **duplicate** natural-key row, a single staging row (one `page_request_id`) matches 2+ dimension rows and fans out into 2+ candidate rows that share the same `page_request_id` but differ in a lookup id. `SELECT DISTINCT` cannot collapse them, and the `INSERT` then writes the same `page_request_id` more than once → violates the unique `IX_PageRequestID`.

The existing `left join hits ... where page_request_id is null` anti-join only prevents **re-importing** ids already in `hits`; it does nothing about duplicates generated **within** the single INSERT by fan-out. The C# `HashSet` dedup in `PageViewsSaveExtension` only dedups staging rows, not the post-join fan-out.

## Impact
The per-day `try/catch` in `AppInsightsImporter.ImportAndSave` logs and **skips the whole day** — every hit for that day is dropped, not just the duplicate. Because the importer rewinds to the newest hit each run, the duplicate dimension keeps tripping current days too → ongoing data loss until the duplicate is removed. `CleanDuplicateUrls` is never called in production and would not cover duplicate sessions/titles anyway.

[...]

PR #359, opened 2026-08-28
Make dbo.urls.full_url unique so duplicate URL lookups can't be created

sambetts:
Makes `dbo.urls.full_url` UNIQUE so duplicate URL lookups can no longer be created — the source-level prevention that complements the defensive fix in #165.

Addresses #167

## Why

`IX_urls_full_url` has been **non-unique** since `ShrinkUrlsFullUrlColumn` created it, so duplicate URLs accumulate in customer databases. Those duplicates are what fan a single page-view out into several hit rows sharing one `page_request_id`, which used to break the unique `IX_PageRequestID` and abort a whole day's import. #165 made the hits merge tolerate that; this stops the duplicates being created.

`IGNORE_DUP_KEY = ON` is the importer-hardening half: a concurrent check-then-insert race then skips only the duplicate URL (as a warning) instead of aborting the whole `INSERT`, so every other new URL in the same batch still lands.

All five production `INSERT INTO urls(full_url)` sites use check-then-insert and then re-join on `full_url`; none reads `SCOPE_IDENTITY()` or `@@ROWCOUNT` from that insert, so a silently skipped duplicate is harmless. No production code path inserts a `Url` via EF, so the "0 rows affected ⇒ `DbUpdateConcurrencyException`" hazard does not reach the importer.

## Migration id, ordering and snapshot

Migration is `202609101100001_UniqueUrlsFullUrlIndex`, sorting immediately after `202609101000001_RetireUnusedAuditYammerStreamTables`.

The **EF entity model is unchanged** — index uniqueness is physical, and this index is created by raw SQL in `ShrinkUrlsFullUrlColumn`, not by the model — so the `.resx` snapshot is a byte-identical copy of its predecessor's, and the manual script stamps `__MigrationHistory` by copying the predecessor's row.

## The part that would have broken upgrades

Two child tables carry a UNIQUE index that includes `url_id`:

| Table | Unique index |
|---|---|
| `file_metadata_property_values` | `IX_url_id_field_id (url_id, field_id)` |
| `hits_clicked_elements` | `IX_hit_id_url_id_timestamp (hit_id, url_id, timestamp)` |

Repointing two duplicate `url_id`s onto one canonical id **collides inside those tables**, so a naive repoint fails with a duplicate-key violation — breaking the upgrade for exactly the customers who have duplicates. Colliding rows are therefore pruned first. Pruning is driven from `sys.indexes` and references from `sys.foreign_keys`, plus the legacy non-FK `event_meta_sharepoint.url_id`.

## Safety — this migration deletes rows

The manual script validates everything up front and aborts with `SET NOEXEC ON` having changed nothing: `__MigrationHistory` exists, the predecessor is stamped, `dbo.urls` exists, and `full_url` is already `nvarchar(850) NOT NULL`.

**Concurrency is now enforced, not just warned about.** Write-intent locks held by another session abort the run *before anything is modified*; the check needs `VIEW DATABASE STATE` and degrades to a warning if that is not granted. Recent import-log rows only **warn** — they cannot prove a live writer, and making that fatal would block legitimate upgrades and repeat the `DenormaliseCopilotChatUserAndTime` mistake of failing a migration over data state. Override with `EXEC sp_set_session_context N'UniqueUrlsFullUrlIndex_SkipConcurrencyCheck', 1`.

**The manual script now proves the schema work landed before stamping.** `sqlcmd`/SSMS continue past a severity-16 error, so a failed `CREATE UNIQUE INDEX` would previously still have reached the stamp and recorded the database as migrated with no unique constraint. A POST-FLIGHT guard checks the index exists, is unique and has `IGNORE_DUP_KEY`. This is a **schema** check only — never a data-state check — and it lives outside the embedded `Up_Sql` region, which a test asserts is verbatim.

## Measured upgrade cost

Synthetic data at scale; no customer database was used. Local SQL Server, `EngineEdition` 2 (Standard), so every figure is the **offline** index build — the worst case, and what Standard/Express customers get. Enterprise / Azure SQL DB / MI take the ONLINE path.

| urls rows | duplicates | total | index rebuild | repoint `hits` | delete |
|---|---|---|---|---|---|
| 1M | 0 (clean) | 13s | 11s | – | – |
| 1M | 20k (2%) | 54s | 10s | 12s | 20s |
| 1M | 200k (20%) | 78s | 8s | 23s | 22s |
| 5M | 100k (2%) | 327s | 59s | 88s | 94s |

- On a database with **no duplicates** the cost is essentially the index rebuild alone (~11s per million urls), because the de-duplication short-circuits after one `GROUP BY`.
- Where duplicates exist the de-duplication dominates, and it scales with the number of **referencing rows repointed** (`dbo.hits` is much the largest), not with the number of duplicate urls: ten times as many duplicates moved the delete only 20s → 22s, because that phase is dominated by per-batch cascade checks across the six child tables.
- Index rebuild is roughly linear (11s/M at 1M, 12s/M at 5M).

Every scale asserted: index unique with `IGNORE_DUP_KEY`, no duplicates remain, no orphaned `hits`, no orphaned legacy `event_meta_sharepoint`, and `hits` row count preserved (repointed, not deleted).

This migration is **correctness-motivated, not performance-motivated**, so the schema-change policy's before/after benchmark gate does not apply. The table above is an **upgrade-cost** measurement for the release notes' maintenance-window estimate. Classify it as a **row-removing** migration.

## Reviewer hotspots from the original PR — all closed

1. **Concurrency** — now enforced (above).
2. **Retention when pruning collisions** — where the table has an `updated` column the most recently updated row now survives, so the freshest `field_value` wins; ties still fall back to already-canonical, then lowest PK, so the result stays deterministic.
3. **Cross-schema FKs** — discovery now carries `OBJECT_SCHEMA_NAME` through and quotes both parts instead of re-qualifying everything as `dbo`.
4. **Filtered / disabled unique indexes** — excluded from collision pruning, so a filtered index cannot drive deletion of rows it does not constrain.
5. **"Already applied" guard** — now also requires `IGNORE_DUP_KEY = ON`, so a hand-created plain UNIQUE index is rebuilt with the correct options rather than accepted as done.

## Test-isolation defect found while validating

`ShrinkUrlsFullUrlColumnMigrationTests` runs against the **shared** unit-test database and its `RestoreMigratedStateAsync` put the schema back to `ShrinkUrlsFullUrlColumn`'s state — which creates `IX_urls_full_url` as **non-unique**. That was the fully-migrated state when the class was written; it no longer is. The result was order-dependent failures in three tests that rely on the new constraint (`DuplicateUrl_CannotBeCreated`, `UrlsDifferingOnlyByCase_AreTheSameUrl`, `EntityTests.URLsTest`), each failing with "Expected 1, Actual 2".

The restore now applies `UniqueUrlsFullUrlIndex.Up_Sql` as well, sending the session-context opt-out **in the same batch** because session context is per-connection and EF hands each command a pooled connection.

## Other changes

- **`EntityTests.URLsTest` revived.** It had been `[Obsolete]` with no `[TestMethod]` since `full_url` was `nvarchar(max)` and could not be indexed. It now asserts the duplicate insert fails, and uses a per-run unique URL plus cleanup because it shares a database.
- **Stale comment corrected** in `Migrate Hits Import into Hits.sql`, which claimed `urls.full_url is not unique`. The `row_number()` fan-out defence stays — `sessions.ai_session_id` is still deliberately non-unique.
- **`CopilotChatFactory.GetOrCreateUrl`** compared the EF `Local` cache ordinally while the database collation is case-insensitive, so a case-only variant could be added twice and then be silently skipped by `IGNORE_DUP_KEY`. Now `OrdinalIgnoreCase` for the in-memory lookup; the translated query is left alone.
- `CleanDuplicateUrls` still has no production caller and remains a one-time safeguard for pre-migration databases.

## Tests

Full suite: **1972/1988 passing**. The 12 failures are pre-existing and environmental — four need a local install-test config file, eight need real Azure/Graph credentials — and none touch URLs or migrations.

Directly relevant: 54/54 across `MigrationSnapshotTests`, `UniqueUrlsFullUrlIndexMigrationTests` (9), `UrlFullUrlMigrationPipelineTests`, `ShrinkUrlsFullUrlColumnMigrationTests`, `DuplicateUrlTests`, `EntityTests`, `HitImportFanoutTests` and `RetireUnusedAuditYammerStreamTablesMigrationTests` — including the previously order-dependent trio, deliberately run in the order that used to break them.

`LatestMigrationSnapshot_IsUpToDateWithTheEntityModel` passes, which is the gate proving the snapshot choice; `ManualScriptUsesTheSameSqlAsTheMigration` proves the POST-FLIGHT guard was added outside the verbatim region.

## Release notes / follow-up

This is the **fifth** migration in the current `dev`→`main` diff. The manual scripts form a prerequisite chain and must be attached to the stable release and run in migration-id order: `AgentCostTables` → `AgentCostUserCredits` → `DlpCopilotImpact` → `RetireUnusedAuditYammerStreamTables` → **`UniqueUrlsFullUrlIndex`** (last). Admin action: stop the importer, take a backup, run in a maintenance window. `Down` only relaxes the index back to non-unique — it **cannot** restore deleted rows.


sambetts:
### Rebased onto current `dev` (post-1810)

This branch was 7 commits behind and conflicted with the two migrations that shipped in **Stable build 1810**. Rebased and resolved — but three of the changes are more than conflict resolution and are worth a look, because a plain "accept both sides" resolution would have shipped a broken migration.

**1. The migration had to be renumbered.**

`202608280900001` (28 Aug) sorted *before* both migrations released in 1810 (`202608310747353`, `202608310800001`), which the repo requires not to happen — a new migration id must sort after the current latest. Renamed throughout to:

```
202608280900001_UniqueUrlsFullUrlIndex  ->  202609010900001_UniqueUrlsFullUrlIndex
```

All four files (`.cs`, `.Designer.cs`, `.resx`, `.manual.sql`), the `IMigrationMetadata.Id`, both `Entities.csproj` item groups, the `__MigrationHistory` guard/stamp in the manual script, and the test that reads the script by name.

**2. The `.resx` model snapshot was stale, and would have caused data loss.**

This is the important one. The migration is raw-SQL only, so by convention its snapshot is a verbatim copy of its predecessor's. Its predecessor *was* `202608250900001_IndexCopilotAccessedResourceFkColumns` — but 1810's `DenormaliseCopilotChatUserAndTime` **changed the EF model** (it added `user_id` / `time_stamp` to `copilot_chats`).

So the carried-over snapshot described a model without those columns:

```
predecessor (ColumnstoreUsageReportMetrics) Target : 59,116 chars
this migration's Target (stale)                    : 59,052 chars
```

Left as-is, EF would have compared the current model against that older snapshot, decided the two columns were surplus, and auto-generated a migration to **drop them** — `AutomaticDataLossException` at best, silent loss of the 1810 denormalisation at worst. The `.resx` is now a byte-identical copy of `202608310800001_ColumnstoreUsageReportMetrics.resx` (verified equal).

**3. Predecessor references updated.** The manual script's prerequisite guard and its `__MigrationHistory` stamp (which copies the predecessor's `Model` blob) now point at `202608310800001_ColumnstoreUsageReportMetrics`, so the chain is `…FkColumns → Denormalise → Columnstore → UniqueUrlsFullUrlIndex` and the stamped `Model` matches the shipped schema.

**Also:** `LatestId` in `UrlFullUrlMigrationPipelineTests` now points here again (it had moved to `ColumnstoreUsageReportMetrics` in 1810).

### Verification

- Release build: **0 errors**.
- This PR's own test set: **16/16 pass**.
- Combined with 1810's migration tests: **21/21**, across 4 consecutive runs.
- ⚠️ **One flake worth knowing about.** On the *very first* run after a fresh build, `DuplicateUrl_CannotBeCreated` and `UrlsDifferingOnlyByCase_AreTheSameUrl` failed, then passed on all four subsequent runs. Both use the shared `AnalyticsEntitiesContext` test database rather than a `ScratchDatabase`, so they appear to depend on that database already being migrated — which the migration-pipeline tests do as a side effect. It is an ordering dependency in the existing suite rather than anything introduced here, but it could surface as an intermittent CI failure and is worth tracking separately.

### Still unresolved (unchanged by the rebase)

This PR was originally reverted once (#253) for **silent data loss**, and that concern is a property of the change itself, not of the rebase: de-duplicating `dbo.urls` means re-pointing **6 FK children** — `hits`, `hits_clicked_elements`, `file_metadata_property_values`, `page_comments`, `page_likes`, `copilot_event_files` — with `hits` being one of the largest tables in the schema. That, plus the deletion being irreversible, is the substance a reviewer should focus on.

sambetts:
This should be left for now. Not a priority yet.

sambetts:
**Release triage 2026-09-03:** #167 is deferred from the current bug-fix release. This PR contains an EF migration for `dbo.urls.full_url`; because the earlier attempt was reverted from `dev` in #253 after benchmarking showed no measurable benefit, the PR needs either a fresh synthetic-scale before/after benchmark plus a dedupe/backfill plan, or it should be closed. I am not closing it as part of this metadata pass.

Issue #360, opened 2026-08-28
Copilot Adoption: slow queries time out, and the page presents the resulting empty data as real figures


sambetts:
## Summary

On a large customer tenant the **Copilot Adoption** page is effectively unusable: it takes minutes to load, frequently returns a `500`, and — more seriously — **when it does render, the figures are wrong**.

App Insights shows five of the eight SQL steps sitting at or near the 90-second per-query timeout. Because every query failure is deliberately degraded to a warning rather than an error, the page renders a normal-looking dashboard built from empty result sets.

This issue covers making that failure honest and survivable. The underlying query tuning is called out separately at the end.

## Evidence

From the `CopilotAdoptionAnalysis` custom event (durations only — no customer data):

```kql
customEvents
| where name == "CopilotAdoptionAnalysis"
| extend TimedOut = tostring(customDimensions.TimedOut),
         Outcome  = tostring(customDimensions.Outcome),
         Slowest  = tostring(customDimensions.SlowestStep),
         TotalMs  = todouble(customMeasurements.TotalMs)
| project timestamp, TotalMs, Outcome, TimedOut, Slowest
```

| timestamp (UTC) | TotalMs | Outcome | TimedOut | SlowestStep |
|---|---:|---|---|---|
| 13:19:20 | 359,773 | Degraded | true | LicenceOpportunities |
| 13:05:51 | 153,889 | Degraded | true | LicenceOpportunities |
| 12:57:53 | 441,735 | Degraded | true | LicenceOpportunities |

Per step (`p50` / `p95` / `max`, ms):

| Step | p50 | p95 | max | Verdict |
|---|---:|---:|---:|---|
| `LicenceOpportunities` | 90,594 | 91,578 | 91,578 | **always times out** |
| `LicensedUsers` | 90,034 | 90,105 | 90,105 | **always times out** |
| `AgentEstate` | 46,180 | 90,011 | 90,011 | slow; sometimes times out |
| `UnlicensedPopulation` | 42,122 | 82,934 | 82,934 | slow |
| `WeeklyTrend` | 3,493 | 90,034 | 90,034 | usually fine; sometimes times out |
| `UsageByApp` | 1,994 | 81,410 | 81,410 | usually fine; sometimes very slow |
| `ResourceTypes` | 654 | 663 | 663 | fine |
| `Scoring` | 49 | 115 | 115 | fine |

`Scoring` at 49 ms shows the in-memory work is healthy — this is entirely SQL.

[...]

PR #383, opened 2026-09-01
Add a Copilot Adoption performance test for before/after measurement

sambetts:
Adds a repeatable way to measure how long the Copilot Adoption page actually takes, so the next round of optimisation can be judged rather than guessed at.

## Why

Release 1810 fixed the two slowest Copilot Adoption queries, and the page is still slow on a large tenant — it spends minutes on *"Analysing Copilot adoption…"*. Going further needs a number that can be produced **before** a change and again **after** it, on the same data, without a customer in the loop.

There was no such number. #360's figures came from a customer's App Insights, and the 1810 benchmarks were hand-written SQL scripts measuring individual queries rather than the page.

## What it does

`Tests.FakeDataGen --run copilotadoption` (or the menu entry) runs the **real** `CopilotAdoptionService` end to end against whatever database the connection string points at, and reports the per-step breakdown.

**It drives the real service on purpose.** Copying the SQL into a benchmark script measures a snapshot that rots silently the moment someone edits `CopilotAdoptionSql`, and it would miss the C# `Scoring` step entirely.

**It reports the instrumentation that already exists.** `CopilotAdoptionService.TimedAsync` already records every step into `CopilotAdoptionDiagnostics` — the same data the page and App Insights use. Step names are compile-time constants and every value is a duration or a count, so **the output carries no tenant data** and is safe to paste into a PR or an issue.

Specifics:

- **Medians** over N timed runs with a cold run done first and discarded — the discipline the SQL benchmarks in this repo already follow.
- **Every window the page offers** (7/28/90/180) by default. The window drives how much of `copilot_chats` each query reads, so a fix that only helps one window is not a fix.
- **Flags any step at ≥95% of `QueryTimeoutSecs`.** Such a step almost certainly *failed* and returned nothing rather than genuinely taking that long. A fast run made of failed steps is not an improvement — that was the entire substance of #360, so the harness refuses to report it as a pass.
- **Optional concurrent callers**, to prove the web layer's shared-analysis cache still de-duplicates. If the measured time scales with the caller count, that de-duplication has broken.
- **JSON output + baseline diff.** Given `COPILOTPERF_BASELINE` it diffs per window and per step against an earlier run and **fails** if the total regressed beyond a tolerance — so it can *gate* a change, not merely describe it.
- **Read-only, generates nothing.** Point it at a database seeded by the existing `UserActivityStressTest`, which already scales to a large tenant.
- **Fully env-var driven** (`COPILOTPERF_REPEATS`, `_ALLWINDOWS`, `_CONCURRENT`, `_OUTDIR`, `_BASELINE`, `_TOLERANCE`, plus `STRESS_NONINTERACTIVE`), so before/after runs are repeatable and scriptable — which is exactly what `BaseStressTest` was built for.

## First baseline, and a finding

Synthetic large-tenant bench, 28-day window. Total **22,811 ms** (cold run 28,006 ms):

| Step | Median ms | Share |
|---|---|---|
| **WeeklyTrend** | **7,683** | **33.7%** |
| LicensedUsers | 7,506 | 32.9% |
| LicenceOpportunities | 3,578 | 15.7% |
| UnlicensedPopulation | 2,569 | 11.3% |
| AgentEstate | 969 | 4.2% |
| Scoring | 277 | 1.2% |
| UsageByApp | 138 | 0.6% |
| ResourceTypes | 9 | 0.0% |

**`WeeklyTrend` is now the slowest step, and 1810 never touched it.** `LicenceOpportunities` — the step whose ~90 s production timeout motivated the columnstore migration — has dropped to fourth. The columnstore work did its job; the next bottleneck is somewhere nobody has looked yet.

The steps run **sequentially**, so the total is their sum: `WeeklyTrend` + `LicensedUsers` are about **two thirds of the wait**.

## Reviewer notes

1. **Only 8 steps report timings, but the service describes ten.** `LicenceTypes` and `DataSourceProbes` do not appear in `Diagnostics.Steps`. Either they are genuinely untimed — in which case the totals above *under-report* — or they are recorded elsewhere. Worth confirming before anyone leans on the shares.
2. **These absolutes are not a production prediction.** The bench runs the whole analysis in ~23 s where production takes minutes: different hardware, no tier cap, no concurrent importer, and a `copilot_chats` row that is materially narrower than the customer's. The *ratio* between runs is the usable signal, which is why the baseline diff compares runs rather than asserting a target.
3. `MemoryMonitor` exposes `InitialMemoryBytes` / `PeakMemoryBytes` / `CurrentMemoryBytes`, and `QueryTimeoutSecs` is a const on `CopilotAdoptionService`, not on `CopilotAdoptionOptions` — both caught by the compiler while wiring this up.

## Testing

Builds clean, and verified end to end against a synthetic large-tenant database — the table above is its real output, not a mock-up.

No product code changes: this only adds a test project file plus its registration in `Program.cs` and the csproj.

### DIFF
--- a/src/AnalyticsEngine/Tests.FakeDataGen/Program.cs
+++ b/src/AnalyticsEngine/Tests.FakeDataGen/Program.cs
@@ -40,6 +40,8 @@ namespace Tests.FakeDataGen
                 ctx => RunStressTest(new ActivityApiDbStressTest(), ctx)),
             new MenuItem("Copilot event import stress test", MenuCategory.StressTest,
                 ctx => RunStressTest(new CopilotStressTest(), ctx)),
+            new MenuItem("Copilot Adoption page performance test (read-only, before/after)", MenuCategory.StressTest,
+                ctx => RunStressTest(new CopilotAdoptionPerfTest(), ctx)),
             new MenuItem("Power Platform event import stress test", MenuCategory.StressTest,
                 ctx => RunStressTest(new PowerPlatformStressTest(), ctx)),
             new MenuItem("Sent email importer stress test", MenuCategory.StressTest,
@@ -76,7 +78,7 @@ namespace Tests.FakeDataGen
             {
                 Console.WriteLine("No SQL connection string provided.");
                 Console.WriteLine("Stress tests that don't need SQL will still run; everything else will be disabled.");
-                Console.WriteLine("Usage: Tests.FakeDataGen.exe \"<SQL Connection String>\" [--run <copilot|activityapi|activityapidb|powerplatform|sentemail|useractivity>]");
+                Console.WriteLine("Usage: Tests.FakeDataGen.exe \"<SQL Connection String>\" [--run <copilot|copilotadoption|activityapi|activityapidb|powerplatform|sentemail|useractivity>]");
             }
             Console.WriteLine();
 
@@ -227,6 +229,7 @@ namespace Tests.FakeDataGen
             new Dictionary<string, Func<BaseStressTest>>(StringComparer.OrdinalIgnoreCase)
             {
                 { "copilot", () => new CopilotStressTest() },
+                { "copilotadoption", () => new CopilotAdoptionPerfTest() },
                 { "activityapi", () => new ActivityAPIStressTest() },
                 { "activityapidb", () => new ActivityApiDbStressTest() },
                 { "powerplatform", () => new PowerPlatformStressTest() },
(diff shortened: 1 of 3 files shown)


sambetts:
Closing as **superseded** - `dev` already carries a strict superset of this branch, and merging it now would *regress* the harness.

`Tests.FakeDataGen/StressTests/CopilotAdoptionPerfTest.cs` on `dev` is the evolved version of the file in this PR. Compared with this branch it additionally has:

- `COPILOTPERF_MAXSTEPS` - makes the step concurrency measurable rather than asserted.
- `NoMigrateContextFactory` - the named, non-migrating `IAnalyticsDbContextFactory`. This branch used an inline context instead. It matters: `ConnectionStringAnalyticsDbContextFactory` passes `autoUpdate: true`, so a measurement harness pointed at a restored copy of a customer database would silently start migrating the schema it is supposed to be measuring.

`Program.cs` and `Tests.FakeDataGen.csproj` are already identical on `dev`. Everything this PR set out to add - real-service (not copied-SQL) measurement, per-step medians with a discarded cold run, every window, the >=95% `QueryTimeoutSecs` failed-step guard, optional concurrent callers, and JSON output with `COPILOTPERF_BASELINE` diffing - is on `dev` today.

Nothing is lost by closing this.

Issue #381, opened 2026-09-01
Tracking: decouple import logic from SQL/REST implementations (ports & adapters)


sambetts:
Tracking issue for a structural initiative: **decouple import logic from concrete SQL and REST implementations**, so that business rules can be unit tested without a database, a Graph tenant, Redis or Service Bus.

This is *Dependency Inversion* applied as *ports & adapters*: logic depends on interfaces (ports), and SQL / Graph / REST / cache are pluggable implementations (adapters) wired up at the composition root.

**This is not a defect report.** Nothing listed here is known to be producing wrong output today. The objective is cleaner separation of concerns and materially better unit-test coverage.

## The headline finding

An audit of every import role in the solution found a consistent asymmetry:

> **The read side is well abstracted. The write side is barely abstracted at all.**

Almost every importer already depends on an `I…SourceLoader` for its Graph/REST reads — and then writes straight to `new AnalyticsEntitiesContext()`, `InsertBatch<T>(connectionString)` or a raw `SqlConnection`. Closing that gap is the substance of this initiative.

## Coverage by import role

| Import role | Read seam | Write seam | Unit tests | Issue |
|---|---|---|---|---|
| Usage-stats upload (`StatsUploader/`) | yes | yes | yes | — already correct |
| Cadence / watermark / delta stores | yes | yes | yes | — already correct |
| Audit log — Activity API | yes | port exists, adapter is a ~550-line class | yes | #373 |
| Copilot audit events | yes | no — ctor takes a connection string | no | #367 |
| Power Platform audit events | n/a | no — ctor takes a connection string | logic only | #367 |
| Copilot usage reports | yes | no — creates its own context mid-flow | DB-backed | #370 |
| Sent email | yes | ad-hoc `Func<DbContext>` | yes | #368 |
| Copilot interaction history | yes | ad-hoc `Func<DbContext>` | yes | #368 |
| Graph users / licences | yes | no — EF + inline `SqlBulkCopy` + raw SQL | DB-backed | #371, #372 |
| Graph usage reports (daily/weekly) | partial | no — EF + `ExecuteSqlCommandAsync` | no | #375 |
| Teams / channels | no | no | no | #377 |
| Calls (Service Bus + webhook) | no — static singleton | no | no | #378 |
| App Insights importer (whole project) | no | no | **none at all** | #369, #374 |
| `GraphImporter` (composition root) | no — `new`s every collaborator | n/a | no | #376 |
[...]

Issue #380, opened 2026-09-01
Parameterise the org_urls insert and drop ToLower() from the EF predicate


sambetts:
Part of the ports & adapters initiative (see the tracking issue #381). Found incidentally during the audit.

## Why

`EnsureOrgURLs` builds a SQL statement by string interpolation of a value that originates from installer configuration, and filters with `.ToLower()` on both sides of the comparison. Neither is a functional defect today, but both are structurally wrong and one of them is a SQL-injection shape that should not exist in the codebase.

Filed as code correctness rather than a defect: current callers supply trusted configuration values, so there is no known exploit path — but the construction is unsafe by pattern and should not be copied.

## Current state (evidence)

`App.ControlPanel.Engine/Models/DatabaseUpgradeInfo.cs:30-34`

```csharp
var existingEntry = db.org_urls.Where(u => u.UrlBase.ToLower() == orgUrl.ToLower()).SingleOrDefault();
if (existingEntry == null)
{
    // We have to manually insert for now, as there's no "org_id" in the model. Not pushing another EF migration just for this.
    db.Database.ExecuteSqlCommand($"insert into org_urls([url_base], [org_id]) values ('{orgUrl.ToLower()}', 1)");
}
```

Two problems:

1. **`:34` — unparameterised SQL.** `orgUrl` is interpolated straight into the statement. A URL containing an apostrophe breaks the statement outright; a hostile value would be executed. `ExecuteSqlCommand` supports parameters and they should be used.
2. **`:30` — `.ToLower()` on both sides of an EF predicate.** EF6 translates this to a SQL `LOWER()` call, which makes the predicate non-SARGable and forces a scan even where an index exists. The database's default collation (`Latin1_General_CI_AS`) is already case-insensitive, so the comparison is redundant as well as costly.

## Proposed change

Parameterise the insert:

```csharp
db.Database.ExecuteSqlCommand(
    "insert into org_urls([url_base], [org_id]) values (@urlBase, @orgId)",
    new SqlParameter("@urlBase", orgUrl),
    new SqlParameter("@orgId", 1));
```

Drop the redundant `.ToLower()` from the predicate and rely on the case-insensitive collation:

```csharp
var existingEntry = db.org_urls.SingleOrDefault(u => u.UrlBase == orgUrl);
```

[...]

PR #384, opened 2026-09-01
Parameterise the org_urls insert and move it behind IOrgUrlStore

sambetts:
Addresses #380. First change in the ports & adapters initiative (#381).

## Scope

`DatabaseUpgradeInfo.EnsureOrgURLs` built its `INSERT` by string interpolation and filtered with `.ToLower()` on both sides of an EF6 predicate:

```csharp
var existingEntry = db.org_urls.Where(u => u.UrlBase.ToLower() == orgUrl.ToLower()).SingleOrDefault();
if (existingEntry == null)
    db.Database.ExecuteSqlCommand($"insert into org_urls([url_base], [org_id]) values ('{orgUrl.ToLower()}', 1)");
```

Neither produced wrong output for trusted installer config, but the interpolation is a SQL-injection shape and `LOWER([url_base])` prevents the predicate seeking `IX_org_urls`.

No schema change, no migration, no behavioural change at any call site. `org_id` is still not on the `OrgUrl` entity, so the insert remains raw SQL - it is now parameterised.

## Changes

| File | Change |
|---|---|
| `App.ControlPanel.Engine/Models/OrgUrlStore.cs` | **New.** `IOrgUrlStore` port, `OrgUrlRules` (pure normalisation), `SqlOrgUrlStore` adapter |
| `App.ControlPanel.Engine/Models/DatabaseUpgradeInfo.cs` | `EnsureOrgURLs` is now pure decision logic over the port; the `AnalyticsEntitiesContext` overload is a thin wrapper |
| `Tests.UnitTests/FakeLoaderClasses/FakeOrgUrlStore.cs` | **New.** Hand-rolled in-memory fake (no mocking library in this solution) |
| `Tests.UnitTests/OrgUrlStoreTests.cs` | **New.** 16 in-memory + 6 DB-backed tests |

`DatabaseUpgrader.CheckDbUpgraded` and every other call site are untouched.

## Design decisions

**The port is synchronous.** The whole call chain is: `CheckDbUpgraded` is a `static void`, and its callers are `Program.cs` (`--initdb`), `DatabaseUpgradeForm`'s `BackgroundWorker.DoWork` handler, `Tests.FakeDataGen` and the unit tests. Async would force a `.Wait()`/`.Result` at that boundary, or a rewrite of the whole chain, for two statements that run once per install. (#380 sketched an async port.)

**Lower-casing on write is kept** - it is load-bearing, not incidental, so #380's suggested `new SqlParameter("@urlBase", orgUrl)` would have been a regression. `Web/AllowCorsForOrgUrlsAttribute.cs` pushes these values straight into `CorsPolicy.Origins`, and in `System.Web.Cors`:

- `CorsPolicy` ctor: `Origins = new List<string>()`, typed `IList<string>`
- `CorsEngine.TryValidateOrigin`: `policy.Origins.Contains(requestContext.Origin)`

`List<string>.Contains` uses `EqualityComparer<string>.Default` - **ordinal, case-sensitive**. A stored `https://Contoso.sharepoint.com` would never match the lower-cased origin a browser sends. Normalisation is now applied **once in C#**, which is what lets the predicate drop its `ToLower()`, and uses `ToLowerInvariant` (under `tr-TR`, `"FINANCE".ToLower()` yields a dotless i and would be stored unmatchable - covered by a test). The other consumer, the importer whitelist (`SiteFilterLoader` / `UrlFilterProcessor.UrlInScope`), lower-cases both sides at match time and is unaffected either way.

## Which guard does what

`dbo.org_urls` carries a **unique** index on `url_base` (`IX_org_urls`, `IGNORE_DUP_KEY = OFF`). A duplicate `INSERT` raises 2627/2601, which propagates to the `catch` in `CheckDbUpgraded` that logs and `return`s - silently abandoning every remaining URL.

The guard that prevents this is the **case-insensitive `Exists()`**: `ExecuteSqlCommand` commits each row before returning, so a later `Exists()` already sees the earlier one, and a row stored in a different case is matched by the collation. The in-config `HashSet` only saves a redundant round trip and cannot help with a pre-existing differently-cased row.

`Any()` replaces `SingleOrDefault()` because this is an existence check: it becomes `IF EXISTS` and never materialises an entity.

## Incidental fix

A null/blank config entry previously threw `NullReferenceException` from `orgUrl.ToLower()`, hitting that same catch-and-return and abandoning every later URL. Blank entries are now skipped.

## Tests

16 in-memory tests run with **zero SQL, Graph, Redis or Service Bus** (#381's rule), covering case-insensitive matching, in-config de-duplication, apostrophes, an injection payload, Greek round-tripping, blank rejection, invariant-culture normalisation and a one-lookup-per-URL scale guard.

6 DB-backed tests are the "genuine round-trip checks" #381 permits and #380 explicitly asks for - parameterisation cannot be proved without a database, since a command-capturing fake would only assert that I wrote the string I wrote.

**Both fixes were tested in reverse:**

| Probe | Result |
|---|---|
| Restore the interpolated insert | apostrophe test fails: `Incorrect syntax near 'brien'` |
| Stub `Exists()` to `false` | differently-cased test fails: `Cannot insert duplicate key row in object 'dbo.org_urls' with unique index 'IX_org_urls'` |

In the second, the reported duplicate key is the *lower-cased* value colliding with the stored *uppercase* row - i.e. the CI collation is genuinely under test.

Full solution builds in Debug; 22/22 pass.

## Reviewer hotspots

- `OrgUrlRules.Normalise` - the lower-case-on-write decision. If you disagree, read the CORS evidence above first.
- `EnsureOrgUrls_AgainstRealDatabase_CaseDifferingConfig_NormalisesToASingleRow` is deliberately named and commented to say it does **not** exercise the collation (C# normalisation settles it before SQL sees either value). `..._PreExistingRowInDifferentCase_IsNotInsertedAgain` is the one that does, which is why it seeds through the store to bypass `Normalise()`.

## Known limitation (pre-existing, out of scope)

A legacy row stored mis-cased (e.g. `HTTPS://CONTOSO.SHAREPOINT.COM`) is reused but **not** rewritten, so ordinal CORS matching still misses it. The old code also only lower-cased on insert, so any installer-created row is already lower-case; repairing existing rows would be a data mutation and a behavioural change, both of which #381 forbids. Worth a follow-up issue if anyone has seen it in the field.

## Review

Reviewed over three rounds by four models (Claude Opus 4.8, GPT-5.6 Sol, Gemini 3.1 Pro, Grok 4.6) until a round came back clean. No defect was ever found in the production code; every finding was an inaccurate justification or a test that did not test what it claimed:

- **R1** - two false claims in my doc comments: that `url_base` had no unique constraint, and that `CheckDbUpgraded` runs on the WinForms UI thread (it runs on `BackgroundWorker.DoWork`). Both corrected.
- **R2** - the dedup comment credited the `HashSet` for a safety property `Exists()` provides; and the case-collision test I had added was tautological (normalisation made both values byte-identical before SQL saw them, so it would have passed on a case-sensitive collation with no unique index). Both corrected.
- **R3** - clean across all four.


sambetts:
The known limitation noted in the description is now filed as #385 (CORS silently fails for org_urls rows not stored in lower case) — pre-existing, not introduced here.

Issue #373, opened 2026-09-01
ActivityReportSqlPersistenceManager: split the adapter into testable collaborators


sambetts:
Part of the ports & adapters initiative for the import engine (see the tracking issue #381).

## Why

The audit-log import is the one role where the **port is already right** — `ActivityImporter<T>` depends only on `IActivityReportLoader<T>`, `IActivitySubscriptionManager`, `IActivityReportPersistenceManager` and `IProcessedBlobStore`, and it has real unit tests using hand-rolled fakes.

The problem is the **adapter**. `ActivityReportSqlPersistenceManager` has grown into a ~550-line class that owns de-duplication policy, Copilot Graph pre-warming, staging-table sharding, concurrency gating, the merge, and the metadata pass. It has no dedicated unit test class, and its collaborators (`SaveSession`, `ActivityImportCache`) construct their own database contexts, so none of its decisions can be observed in isolation.

This is a split, not a new port. `IActivityReportPersistenceManager` stays exactly as it is.

## Current state (evidence)

`WebJob.Office365ActivityImporter.Engine/ActivityAPI/ActivityReportSqlPersistenceManager.cs`
- `:103` `CommitAll` — creates a context purely to read the default connection string.
- `:212` `GetOrBuildRunImportCacheAsync` — run-scoped de-dup cache lifecycle plus the `AUDIT_PERBATCH_DEDUP_CACHE` safety-valve branch.
- `:249` `GetSharedCopilotLoaderAsync` — builds a Graph client (network) inside the persistence adapter.
- `:286` `PrewarmCopilotFileMetadataAsync` — Graph pre-warm concurrency policy.
- `:350` `SaveToSqlAllTheThings` — staging table naming/sharding, the merge, and the concurrency semaphores.
- `:450` `SaveMetadataAsync` — the per-workload metadata pass.

`WebJob.Office365ActivityImporter.Engine/ActivityAPI/SaveSession.cs:26` and `:36` — takes a concrete `AnalyticsEntitiesContext`, and `Init()` constructs `GraphAppIndentityOAuthContext`, `GraphServiceClient`, `GraphFileMetadataLoader`, `CopilotAuditEventManager` and `PowerPlatformAuditEventManager` inline.

`WebJob.Office365ActivityImporter.Engine/ActivityImportCache.cs:36-42` — `static GetAndBuildNewCache(...)` creates `new AnalyticsEntitiesContext()`. A static factory that opens a database is unmockable by construction.

[...]

PR #417, opened 2026-09-02
Split ActivityReportSqlPersistenceManager into testable collaborators (#373 part 2)

sambetts:
Addresses #373 (part 2 of 2). Part of #381.

Part 1 (#396) lifted the pure rules out of `ActivityReportSqlPersistenceManager`. This is **part 2: the collaborator split**, so the adapter's remaining jobs each sit behind a seam that can be exercised without SQL Server or Graph.

**Behaviour is identical.** No change to the SQL, the staging schema, the merge script, the concurrency design, the log text, or the numbers inside it. No schema change, no migration, no index change. `InsertBatch<T>` keeps its row-by-row implementation - it is only reached through a port now.

## The two operator safety-valves

`#373` says this class carries two, and both keep working and are now assertable:

| Valve | App setting | Where it lives now | Test |
|---|---|---|---|
| Per-batch de-dup cache | `AUDIT_PERBATCH_DEDUP_CACHE` | `ActivityImportCacheProvider` | `ImportCache_PerBatchSafetyValve_RebuildsForEveryBatchOverThatBatchsOwnSpan` |
| Concurrent saves | `AUDIT_MAX_CONCURRENT_SAVES` | `ActivitySaveConcurrencyPolicy` (part 1) + `ActivityStagingPass` | `SavePipeline_SerialMode_...` / `SavePipeline_ShardedMode_PointsTheMergeAtTheShardAndHandsDownTheSharedWriteLock` |

## What moved

All new files under `WebJob.Office365ActivityImporter.Engine/ActivityAPI/Persistence/`.

| Was (in the manager) | Is now |
|---|---|
| `ActivityImportCache.GetAndBuildNewCache(...)` (static, `new AnalyticsEntitiesContext()`) | `IActivityImportCacheLoader` + `SqlActivityImportCacheLoader` |
| `GetOrBuildRunImportCacheAsync` + the per-batch/per-run ternary in `CommitAll` | `IActivityImportCacheProvider` + `ActivityImportCacheProvider` |
| `new EFInsertBatch<AuditLogTempEntity>(db, _logger)` + `SaveToStagingTable(...)` | `IActivityStagingWriter` / `IActivityStagingBatch` + `SqlActivityStagingWriter` / `SqlActivityStagingBatch` (internal) |
| `GetSharedCopilotLoaderAsync` (built a `GraphServiceClient` inline) + `PrewarmCopilotFileMetadataAsync` | `ICopilotMetadataLoaderFactory` + `GraphCopilotMetadataLoaderFactory`, and `CopilotMetadataPrewarmer` |
| `new SaveSession(_logger, db, _appConfig, loader)` + `await Init()` | `ISaveSessionFactory` + `SaveSessionFactory` |
| The per-event dedup/scope/stage loop and the staging load + merge inside `SaveToSqlAllTheThings` | `ActivityStagingPass` |

`ActivityStagingPass` is the piece that makes the rest testable: it reaches SQL **only** through `IActivityStagingBatch`, so a batch's `ImportStat` counters, its operator log line, its phase timings and its merge wiring can all be asserted with no database. It is stateless between calls (everything is per-`RunAsync`), so one instance is shared by every batch of a cycle, including concurrent ones.

### Deliberate deviation from #373's proposed interface shape

#373 proposed `IActivityStagingWriter` as three methods - `CreateStagingTableAsync` / `LoadAsync` / `MergeAsync`. It is one method here (`LoadAndMergeAsync`), because of how `InsertBatch.SaveToStagingTable` is built: it opens one connection, creates the staging table on it, fans the row inserts out over a **separate `SqlConnection` per chunk**, then runs the merge back on the original - holding that original connection open throughout. That held-open connection is what keeps the staging table alive and visible to the chunk connections and the merge (in Release the default table is the `##` global temp table; in Debug it is the permanent `debug_import_staging_event_lookups`). Three port calls would either break that lifetime or force the port to hold connection state across calls. Collapsing them keeps the existing SQL byte-identical.

Also, `ActivityDedupPolicy` from the issue's sketch is not added: its two proposed members already exist as `ActivityStagingRules.IsAlreadyProcessed` and `AuditFilterConfig.InScope`, both from part 1.

## Wiring

Both existing public constructors are kept as **delegating overloads** - no call site changes (`ProgramTasks.cs`, `ActivityImporterTests`, `ActivityApiDbStressRunner`, `Tests.UnitTests/Program.cs`). The new collaborator constructor is `internal` because `IActivityStagingWriter` carries the internal `AuditLogTempEntity`.

A `null` collaborator on that constructor means "build the production adapter", constructed in the body rather than in a chained initialiser. To be accurate about what that buys: it changes **nothing** today - none of the production adapters validates or dereferences what it is handed, so a null `appConfig` reaches `new UserGroupsFilterModel(appConfig.UserGroupsFilter)` and raises the same `NullReferenceException` either way. It is deliberate ordering-insurance, so that adding a guard to one of those adapters later cannot silently change which exception an existing caller sees. Happy to swap it for required parameters if reviewers prefer.

Consistent with that, `ActivityStagingPass`'s constructor deliberately has **no** null guards: the manager did not validate `filterConfig` / `userGroupsCache` either, and it dereferences them mid-save. An `ArgumentNullException` at construction where there used to be a `NullReferenceException` mid-save is an operator-visible change, not an extraction.

## Tests - 17 new, zero SQL Server / Graph / Redis / Service Bus

**`ActivityImportCacheProviderTests`** (4)
- `ImportCache_PerRunScope_IsBuiltOnceAndReusedByEveryBatch` - one load per cycle, same instance to every batch, over the whole download window rather than the batch span.
- `ImportCache_PerBatchSafetyValve_RebuildsForEveryBatchOverThatBatchsOwnSpan` - the valve reloads per batch **and** uses each batch's own `[oldest, newest]`.
- `ImportCache_ConcurrentFirstUse_StillBuildsOnlyOnce` - 8 concurrent first callers, one load.
- `ImportCache_PerRunBuild_LogsTheIdCountAndWindowOnce` - the operator-facing "built run dedup cache" line, its `ProcessedIdCount` and its `N-day window`, logged once per cycle.

**`CopilotMetadataPrewarmerTests`** (6)
- `Prewarm_CopilotResourceResolutionDisabled_MakesNoGraphFileLookups` - a tenant that opted out must generate **zero** outbound Copilot file lookups (the loader is still returned, because `SaveSession` gets it either way).
- `Prewarm_CopilotResourceResolutionEnabled_ResolvesEachDistinctContextOnceWithItsOwnUpn` - one round-trip per distinct context, keyed with the first claiming event's UPN, and non-Copilot events cost nothing.
- `Prewarm_LoaderFactoryFails_ReturnsNullAndWarnsAndSkipsThePrewarm` - degrades to null with the exact warning, exception included.
- `Prewarm_LoaderIsBuiltAtMostOncePerCycle_EvenWhenTheFactoryFailed` - a **failed** Graph auth is not retried (nor re-warned) on the next batch.
- `Prewarm_OneContextFailingInGraph_DoesNotStopTheOthers` - swallowed at Debug; no Error/Warning.
- `Prewarm_FanOutIsThrottled_ButStillConcurrent` - peak concurrency `> 1` **and** `<= 8` (the literal cap, deliberately not `PrewarmConcurrency`, so raising that constant cannot pass silently).

**`ActivitySavePipelineTests`** (7)
- `SavePipeline_EachOutcomeIsStagedAndCountedCorrectly` - `Total` / `Imported` / `UsersOutOfScope` / `URLsOutOfScope`, only the qualifying event staged and handed to the metadata pass. It also pins that a de-duplicated event increments **nothing** (`ProcessedAlready` stays 0) - long-standing behaviour, see the note below.
- `SavePipeline_UserOutsideGroupsFilter_LogsTheOperatorLineNamingTheUser` - exact message text.
- `SavePipeline_SerialMode_MergesTheSharedStagingTableWithNoMergeLock` - no table override (so `InsertBatch` uses `[TempTableName]`), no merge lock, `10000` inserts per thread, every `${STAGING_TABLE_ACTIVITY}` substituted.
- `SavePipeline_ShardedMode_PointsTheMergeAtTheShardAndHandsDownTheSharedWriteLock` - the merge reads this save's shard, not the shared table, and gets the same `SemaphoreSlim` instance.
- `SavePipeline_StagingRowCarriesTheEventsOwnUserAndUrlUnchanged` - the row's `user_name` is the event's own UPN, and a Greek SharePoint URL reaches SQL unchanged.
- `SavePipeline_SlowScopeCheck_IsChargedToTheDedupPhaseNotTheMerge` / `SavePipeline_SlowMerge_IsChargedToTheMergePhaseNotTheDedup` - a **crossed pair**, so swapping the two phase timings fails both. A single "is a timing recorded?" test would not.

New fakes in `Tests.UnitTests/FakeLoaderClasses/FakeActivityPersistencePorts.cs`: `InMemoryActivityStagingWriter` / `InMemoryActivityStagingBatch`, `FakeActivityImportCacheLoader`, `FakeCopilotMetadataLoaderFactory`, `RecordingCopilotMetadataLoader`, `PredicateAuditFilterConfig`, `RecordingLogger`. Failure fakes fault their `Task` rather than throwing synchronously, so a dropped `await` cannot hide behind the caller's `catch`.

## Verification

- Full solution builds clean in Debug, no new warnings.
- **74 / 74 pass** across the 17 new tests plus the pre-existing tests for this area: `ActivityImporterTests`, `AuditImportFailureHandlingTests`, `AuditTests`, `ActivityStagingRulesTests`, `ActivitySaveConcurrencyPolicyTests`, `ActivityImportCacheWindowTests`, `CopilotPrewarmExtractionTests`, `ActivityReportSetMinMaxTests`, `ActivityReportLoaderTimeoutTests`. The DB-backed `ActivityImporterTests` drive the **real** manager through full saves against LocalDB, with both real and generated audit data - that is the behaviour-identical evidence. To be precise about what that does and does not cover: every real-manager construction in that class uses the four-argument overload, so `maxConcurrentSaves` is `1` and only the **serial** save path is exercised end-to-end against SQL. The sharded/concurrent path is covered by `AuditImportFailureHandlingTests.ConcurrentSaveMode_...` (real orchestration, fake persistence manager), by `ActivitySaveConcurrencyPolicyTests`, and by the new `SavePipeline_ShardedMode_...`; the real-manager concurrent case is the separately gated stress harness, which is not in this 74-test set. Re-run after rebasing onto the current `dev`.
- **Every one of the 17 new tests was mutation-checked.** Over six rounds the production line each test claims to cover was broken on purpose, and in each round *exactly* the predicted set of tests failed and no others: `&& false` on the per-batch branch; the cache init lock removed; the build-once flag removed; `ShouldPrewarm(..., true)`; `_sharedCopilotLoaderTried = false`; `SemaphoreSlim(1)` and then `SemaphoreSlim(int.MaxValue)` instead of `PrewarmConcurrency`; both `try/catch`es defeated; `GetSpoFileInfo` arguments swapped; `EffectiveStagingTableName(null)`; the effective name passed as the override; the two phase timings swapped; `URLsOutOfScope++` -> `UsersOutOfScope++`; a hard-coded UPN on the staging row; and the skip message reworded. The two tests hardened during review were re-checked afterwards and now fail more decisively than before (removing the cache init lock yields `LoadCount == 8`, not merely `> 1`; an unbounded prewarm yields peak concurrency 24 against the literal cap of 8).

Two tests were tightened as a result of that exercise: one asserted `LastInsertsPerThread` against the very constant that produces it (so changing the constant could not fail it - now a literal `10000`), and a `MakesNoGraphCalls` test was too weak to fail on any realistic break, so it was folded into the dedup test as an extra non-Copilot event rather than kept as its own case.

## Performance at 200k-user scale

No new per-event work and no new per-event allocation. The hoisted `urlInScope` / `userInGroupsFilter` / `stageRow` delegates are still built once per batch, not per event; the staged rows are still added to one list owned by `InsertBatch`, with no intermediate copy; the events handed to the metadata pass are still the same `ConcurrentBag`, so its enumeration order is unchanged. Per save batch the split adds one `IActivityStagingBatch` wrapper object; `ActivityStagingPass`, `ActivityImportCacheProvider`, `CopilotMetadataPrewarmer` and `SaveSessionFactory` are each built once per manager (i.e. once per import cycle).

## Reviewer hotspots

1. **`ActivityStagingPass.RunAsync` vs the old `SaveToSqlAllTheThings`** - the biggest move. Worth reading as a literal diff: the statement order, the `#if DEBUG` line, the stopwatch boundaries and the `if/else if` counter chain are intended to be unchanged.
2. **The `null`-means-production collaborator constructor** - see the reasoning above; happy to change it if the ordering argument does not hold up.
3. **`ISaveSessionFactory` has no DB-free test.** It is honest to say so: the metadata pass still needs a real `AnalyticsEntitiesContext` (`db.AuditEventsCommon` / `db.sharepoint_events`) and `ProcessExtendedProperties` takes the concrete `SaveSession`, so making that pass DB-free is a part-3 change, not this one. The factory is covered by the DB-backed `ActivityImporterTests`.
4. **The double-checked init in `ActivityImportCacheProvider` / `CopilotMetadataPrewarmer`** pre-checks a non-volatile `bool` outside the lock, then returns a separately-published reference. This is the pre-#373 code moved verbatim, so it is not a regression, and I left it alone rather than widen a pure-move PR. Reviewers split on how to describe it: x86/x64 hardware does not reorder store/store or load/load, but the JIT is an independent reordering agent and an ordinary read is not an acquire - so the doc comments now call it implementation behaviour rather than a guarantee, which is the conservative reading. **Say the word and I will add `Volatile.Read`/`Volatile.Write`**, here or in a follow-up.
5. **`CommitAllToSQL` still opens a real `SqlConnection`**, so `CommitAll` itself is not DB-free either. Deliberately left alone - abstracting connection creation is a bigger blast radius than this PR should carry.

## Out of scope, noted while reading

`ImportStat.AddStats` accumulates every counter **except** `UsersOutOfScope`, while `ImportStat.ToString()` prints it - so the per-cycle summary always reports `users out of scope: 0` no matter how many events the groups filter excluded. Pre-existing (it predates part 1) and **not** fixed here, because it changes an operator-facing number, which #381 forbids in this PR. Filed as #418.

## Review

Three `code-review` agents (claude-opus-4.8, gpt-5.6-sol, grok-4.6) ran two rounds against this branch. **No reviewer found a production defect in either round** - all three independently concluded the extraction is behaviour-identical. Every finding was an inaccurate claim or a test-strength issue, and each was verified against the source before being acted on:

**Round 1** - `SavedToSql` doc claimed insertion order (`ConcurrentBag` gives none) and that its events reached SQL (`InsertBatch` skips over-width rows); the constructor rationale claimed an exception change that does not happen today; two `Thread-safe` docs were unqualified; `Prewarm_FanOutIsThrottled` asserted against the same constant that sizes the semaphore; `ImportCache_ConcurrentFirstUse` used `Task.Run` and could be staggered by ThreadPool thread injection; and this PR body overstated concurrent coverage.

**Round 2** - the reworded memory-model claim was still stronger than the CLR guarantees; "`ActivityImportCache` is internally locked" was true only of its de-dup members; "none of the adapters validates what it is handed" was false for two of them; "rows the merge created" ignored that a concurrent batch's metadata pass can match a row the *other* merge inserted; the one-connection staging rationale did not match `InsertBatch` (chunk loads use their own connections); and round 1's raw-`Thread` racer test could terminate the test host on an unhandled exception. All fixed in `6abb733`.

Two remaining nits at the end of round 2 were PR-body staleness, corrected above.

### DIFF
--- a/src/AnalyticsEngine/Tests.UnitTests/ActivityImportCacheProviderTests.cs
+++ b/src/AnalyticsEngine/Tests.UnitTests/ActivityImportCacheProviderTests.cs
@@ -1,0 +1,136 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Tests.UnitTests.FakeLoaderClasses;
+using WebJob.Office365ActivityImporter.Engine;
+using WebJob.Office365ActivityImporter.Engine.ActivityAPI.Persistence;
+using WebJob.Office365ActivityImporter.Engine.ActivityAPI.Rules;
+using WebJob.Office365ActivityImporter.Engine.Entities;
+using WebJob.Office365ActivityImporter.Engine.Entities.Serialisation;
+
+namespace Tests.UnitTests
+{
+    /// <summary>
+    /// The audit-log de-duplication cache lifecycle, split out of ActivityReportSqlPersistenceManager by
+    /// issue #373 part 2. Runs with zero SQL Server: the only thing that touched the database was the load
+    /// itself, and that is now <see cref="IActivityImportCacheLoader"/>.
+    ///
+    /// What is being pinned here is one of the two operator safety-valves this class carries -
+    /// <c>AUDIT_PERBATCH_DEDUP_CACHE</c> - plus the build-once-per-cycle optimisation it exists to undo.
+    /// </summary>
+    [TestClass]
+    public class ActivityImportCacheProviderTests
+    {
+        private static readonly DateTime Now = new DateTime(2026, 3, 1, 9, 30, 0, DateTimeKind.Utc);
+
+        private static AbstractAuditLogContent Event(DateTime createdUtc)
+            => new AzureADAuditLogContent { Id = Guid.NewGuid(), UserId = "someone@contoso.onmicrosoft.com", CreationTime = createdUtc };
+
+        private static ActivityDedupCacheWindow PerRunWindow(int daysBeforeNowToDownload = 3)
+            => ActivityImportCacheWindow.Resolve(usePerBatchDedupCache: false, oldestContentUtc: Now.AddHours(-1),
+                newestContentUtc: Now, daysBeforeNowToDownload: daysBeforeNowToDownload, nowUtc: Now);
+
+        private static ActivityDedupCacheWindow PerBatchWindow(DateTime oldestUtc, DateTime newestUtc)
+            => ActivityImportCacheWindow.Resolve(usePerBatchDedupCache: true, oldestContentUtc: oldestUtc,
+                newestContentUtc: newestUtc, daysBeforeNowToDownload: 3, nowUtc: Now);
+
+        [TestMethod]
+        public async Task ImportCache_PerRunScope_IsBuiltOnceAndReusedByEveryBatch()
+        {
+            var loader = new FakeActivityImportCacheLoader();
+            var provider = new ActivityImportCacheProvider(loader, new RecordingLogger());
+            var window = PerRunWindow();
+
+            var first = await provider.GetForWindowAsync(window);
+            var second = await provider.GetForWindowAsync(window);
+            var third = await provider.GetForWindowAsync(window);
+
+            Assert.AreEqual(1, loader.LoadCount,
+                "The run-scoped cache must be loaded from audit_events ONCE per import cycle, not once per save batch.");
+            Assert.AreSame(first, second, "Every batch of the cycle must share one cache, so ids remembered by one batch are seen by the next.");
+            Assert.AreSame(first, third);
+
+            // ...and over the whole download window, not the batch's own [oldest, newest] span.
+            Assert.AreEqual(window.FromUtc, loader.Loads[0].FromUtc);
+            Assert.AreEqual(window.ToUtc, loader.Loads[0].ToUtc);
+        }
+
+        [TestMethod]
+        public async Task ImportCache_PerBatchSafetyValve_RebuildsForEveryBatchOverThatBatchsOwnSpan()
+        {
+            // AUDIT_PERBATCH_DEDUP_CACHE=true. The valve exists so an operator can restore the pre-#373
+            // per-batch behaviour without a redeploy, so it must genuinely reload for each batch AND use
+            // each batch's own span - not the cycle-wide window.
+            var loader = new FakeActivityImportCacheLoader();
+            var provider = new ActivityImportCacheProvider(loader, new RecordingLogger());
+
+            var batchOne = PerBatchWindow(Now.AddHours(-5), Now.AddHours(-4));
+            var batchTwo = PerBatchWindow(Now.AddHours(-2), Now.AddHours(-1));
+
+            var first = await provider.GetForWindowAsync(batchOne);
+            var second = await provider.GetForWindowAsync(batchTwo);
+
+            Assert.AreEqual(2, loader.LoadCount, "With the safety-valve on, every batch rebuilds the cache.");
+            Assert.AreNotSame(first, second, "Each batch gets its own cache instance under the safety-valve.");
+
+            Assert.AreEqual(Now.AddHours(-5), loader.Loads[0].FromUtc);
+            Assert.AreEqual(Now.AddHours(-4), loader.Loads[0].ToUtc);
+            Assert.AreEqual(Now.AddHours(-2), loader.Loads[1].FromUtc);
+            Assert.AreEqual(Now.AddHours(-1), loader.Loads[1].ToUtc);
+        }
+
+        [TestMethod]
+        public async Task ImportCache_ConcurrentFirstUse_StillBuildsOnlyOnce()
+        {
+            // In concurrent-save mode several batches enter the save path at once, all before the cache
+            // exists. Without the init lock each of them would run the (expensive, whole-window)
+            // audit_events query.
+            var loader = new FakeActivityImportCacheLoader { LoadDuration = TimeSpan.FromMilliseconds(250) };
+            var provider = new ActivityImportCacheProvider(loader, new RecordingLogger());
+            var window = PerRunWindow();
+
+            var racers = Enumerable.Range(0, 8)
+                .Select(_ => Task.Run(async () => await provider.GetForWindowAsync(window)))
+                .ToArray();
+            var caches = await Task.WhenAll(racers);
+
+            Assert.AreEqual(1, loader.LoadCount, "Concurrent first callers must not each build the cache.");
+            Assert.IsTrue(caches.All(c => ReferenceEquals(c, caches[0])), "All concurrent callers must get the same cache instance.");
+        }
+
+        [TestMethod]
+        public async Task ImportCache_PerRunBuild_LogsTheIdCountAndWindowOnce()
+        {
+            // Operator-facing telemetry: #381 forbids changing log text or the numbers inside it, and this
+            // line is how an operator sees whether the per-cycle cache is working.
+            var logger = new RecordingLogger();
+            var loader = new FakeActivityImportCacheLoader
+            {
+                SeedCache = cache =>
+                {
+                    cache.RememberProcessedEvent(Event(Now.AddHours(-1)));
+                    cache.RememberProcessedEvent(Event(Now.AddHours(-2)));
+                    cache.RememberProcessedEvent(Event(Now.AddHours(-3)));
+                }
+            };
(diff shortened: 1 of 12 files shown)


sambetts:
Heads-up: this merged while round 2 of the multi-model review loop was still running, so round 2's fixes are not in it. They are now up as #419 — five over-broad doc claims narrowed (no production behaviour change) plus a host-safety fix to the racer test this PR's round-1 fixes introduced.

Neither review round found a production defect: claude-opus-4.8, gpt-5.6-sol and grok-4.6 all independently concluded the extraction is behaviour-identical.

Issue #437, opened 2026-09-03
Reports: add always-visible licence activity explorer with ranking and search


sambetts:
## Goal

Add a **Licence activity** sub-tab to the in-app Reports page that lets an administrator explore user activity by assigned Microsoft 365 licence.

## Required UX

- Keep the tab visible regardless of import configuration.
- When `GraphUsersMetadata` is disabled, show an informative empty state explaining that the user metadata import supplies licence-to-user assignments and must be enabled. Do not hide the tab.
- When user metadata is enabled but an activity source is disabled or has not imported data yet, identify the affected workload and explain why its estimate is unavailable. Do not render unavailable activity as zero.
- Show every imported licence SKU and its assigned-user count.
- Let the user select a licence and reporting period.
- Show a configurable X most-active and X least-active users for the selected licence.
- Include users with measured zero activity in the least-active view.
- Include a username/UPN search that runs server-side and does not require loading the full licence population.
- For each user, show separate, clearly labelled activity estimates for Outlook, SharePoint, OneDrive, Teams and Copilot, plus the evidence/source needed to understand each estimate.
- Clearly distinguish `0`, unavailable, and not-yet-imported states.

## Interaction and accessibility

- Use the existing Fluent UI portal patterns.
- Keep ranking controls, search and licence selection keyboard accessible.
- Define deterministic ordering for tied scores.
- Keep result sizes bounded; use paging or incremental loading where search/ranking results can exceed the selected X.
- Keep licence selection and user results responsive with 300,000 users and 20 licence types; never require loading every assignment into the browser.
- Preserve the selected reporting period and sensible filters while switching licences.

## Acceptance criteria

1. **Licence activity** is always present in the Reports tab list.
2. With `GraphUsersMetadata=false`, selecting it shows the prerequisite message.
3. With metadata enabled, the page shows licence counts and supports licence selection.
4. The most/least-active count is configurable and both lists update from the API.
5. Username/UPN search works within the selected licence.
6. Workload cells visually distinguish unavailable data from genuine zero activity.
7. UI tests cover the disabled prerequisite, partial activity-source availability, empty licences, zero-activity users, searching and ranking controls.

sambetts:
**Release triage 2026-09-03:** deferred from the current bug-fix release because this is the portal/reporting half of the licence-activity feature, not a standalone bug fix. It should ship with #436 in the `Reports: licence activity by user` milestone after the scalable per-user workload scoring API is designed and validated.

[...]

Issue #436, opened 2026-09-03
Licence activity reports: add scalable per-user workload scoring API


sambetts:
## Goal

Provide a scalable API/query layer for the Reports page to browse licence assignments and compare each assigned user's activity across Outlook, SharePoint, OneDrive, Teams and Copilot.

The point-in-time licence relationship already exists in `license_types` and `user_license_type_lookups`; do not rebuild licence importing. Reuse the existing Copilot adoption scoring/query primitives where their semantics fit, but keep this API general across all licence SKUs.

## API contract

- Return every imported licence SKU with its assigned-user count.
- Return ranked users for one selected licence and reporting period.
- Support a bounded `top`/`bottom` count and server-side username/UPN search.
- A user with multiple licences belongs to each applicable licence result.
- Preserve users with no matching activity rows through left joins so they receive genuine zero-activity estimates and can appear in the least-active list.
- Include enough raw evidence alongside each estimate to make it explainable, such as active days, workload event/count totals and last-activity date where available.
- Return explicit availability for:
  - `GraphUsersMetadata` disabled: licence-to-user reporting unavailable.
  - An activity import disabled or not yet populated: that workload is unavailable, not zero.
  - Imported activity with no activity for a user: genuine zero.

## Activity sources

- Outlook, SharePoint, OneDrive and Teams should use the existing Graph usage-report tables and the settled-snapshot/reporting-window semantics already used by `ReportsAPIController`.
- Copilot should use the best available per-user source already supported by the deployment, preferring the official Graph Copilot usage report where available and exposing which source informed the estimate.
- Define a transparent, deterministic per-workload activity estimate. Avoid combining unlike raw counts into an unexplained total.

## Scale and correctness

- Design and test for 300,000 users and 20 licence types, including users assigned to multiple licences and therefore potentially millions of user-to-licence result relationships.
- Rank, search and page in SQL; never materialise the whole user-by-licence population in the web process or browser.
- Avoid per-user queries and non-SARGable transformations of indexed username columns.
- Bound all result sizes and validate query behavior for small, common and tenant-wide SKUs across all 20 licence types.
- Add tests for multiple licences, zero activity, unavailable imports, delayed/missing snapshots, username search, deterministic ties, and top/bottom boundaries.
- If a schema/index change is proposed, benchmark the real queries at synthetic scale before including it.

## Acceptance criteria

1. Consumers can list licences and assigned-user counts.
[...]

PR #456, opened 2026-09-05
Licence activity reporting: aggregate explorer and admin drill-down

sambetts:
## What this adds

An always-visible **Licence activity** sub-tab under Reports, answering one question: which licences are assigned, and how much are the people holding them actually using each Microsoft 365 workload.

It reports **distinct current assignments** read from the existing `license_types` / `user_license_type_lookups` tables, where SKUs overlap and one user can hold many. It is deliberately **not** capacity, ownership history, SKU attribution, or a productivity/ROI score. The five workloads (Teams, Outlook, OneDrive, SharePoint, Copilot) are always reported separately and never blended.

**No migration, no schema change, no installer config-schema change** (`CONFIG_VERSION` untouched). Nothing new is imported - this reads what the existing importers already store.

## Merged to dev; stable acceptance remains blocked

This PR was merged into `dev` at `b415bc088a4c7bc3017cc1dfd9c0507121d8e3cc` using the user's explicit admin-override authorisation. That was not approval to merge `main` or to treat performance acceptance as passed. The original LocalDB evidence below is retained as diagnostic history; the subsequent real Azure S4/S6 measurements are recorded at the end of this body and in draft release PR #458. No product schema change, range reduction or production-timeout relaxation was made.

Measured on a synthetic 300,000-user / 50-SKU database (7,800,000 rows in each of the five workload activity tables, 2,060,681 licence assignments), **in both index modes**:

| Scenario | B-tree fallback | Columnstore (production default) | Budget |
|---|---|---|---|
| 7-day adapter overview, first instrumented call | 19,670 ms | **15,203 ms** | HTTP target: 15,000 ms; this row is not an HTTP measurement |
| 7-day adapter overview, warm median (n=3) | 12,721 ms | 10,985 ms | - |
| 7-day logical reads | 1,011,562 | 1,011,562 | - |
| **180-day adapter overview** | **Execution Timeout Expired** | **Execution Timeout Expired** | 20,000 ms command timeout |

The 180-day window is an **offered preset**, and it exceeds the 20-second per-command production timeout (`LicenceActivitySql.CommandTimeoutSeconds`) in both index modes. Columnstore is the production index shape because `202608310800001_ColumnstoreUsageReportMetrics` is already on `dev`. Its observed 7-day first-call elapsed time was lower, but these shared-VM runs do not isolate an index-only causal improvement; the 180-day timeout occurred in both modes.

### Honest caveats on those numbers

These were audited by a separate reviewer specifically to stop me overclaiming, and several of my first-draft claims did not survive:

- This is **LocalDB, not Azure SQL**. Treat it as a relative signal, not a production SLA.
- "Cold" in the table means the **first adapter call in the measurement process**, not an HTTP response, a cold deployed web process, or a cold SQL buffer pool. The harness deliberately does not run `DBCC DROPCLEANBUFFERS`; SQL buffer and plan caches are retained/uncontrolled.
- Cold runs carry `SET STATISTICS IO/TIME` overhead; the warm median does not. The two are not perfectly comparable, and `logicalReads` comes from a separate instrumented call.
- The five workload loads run **concurrently at MAXDOP 1**, so per-operation elapsed **overlaps and is not additive**.
- Under the 600-second test-only diagnostic timeout the 180-day window returns in 106,095 ms, with the heaviest statements at about 37 s CPU vs about 90 s elapsed and about 82k logical reads each. **I first read that as tempdb spilling; that is not supported.** The CPU/elapsed gap alone does not distinguish CPU scheduling, memory-grant waits, I/O, blocking or spilling. Concurrent work can affect elapsed time; root cause and quiet acceptance remain unproven.
- No schema or production timeout was changed. The diagnostic probe explicitly used a 600-second test-only command timeout to collect failure evidence; that path reports **Inconclusive**, never Passed and never acceptance.

## Correctness properties that drove the design

- **Unknown is never rendered as zero.** A licensed user with no activity row is *unknown*; a user with a complete set of explicit zero rows is a *measured zero*, and only the latter may be ranked least-active.
- **Most/least active (1..100 inclusive) are always ranked by the selected workload**; the sort parameter affects the browse list only.
- M365 snapshot counters are **averages over the settled window, never a sum of daily events**. Copilot pins the actual date and period, requires a fully contained D7 window, and treats v1 frequency as unknown.
- Windows are inclusive UTC, 7..180 days ending before today; default 28 days ending the latest settled Sunday. **Custom ranges are never rounded.**
- There is no user displayName in the product, so search is by UPN/mail only and the UI never invents one.

## Access

Aggregates are available to every already-authenticated portal user. Individual detail and individual export require the optional Entra app role **`LicenceActivity.ReadUsers`**; without it the tab degrades to aggregates rather than disappearing. The role is **re-checked at export time**, so a snapshot minted while the role was held cannot be exported after it is revoked. This does not change authorization for any other portal area.

## Boundedness

16 overview snapshots (5 min) and 32 user snapshots (2 min, capped to the parent overview lifetime), four shared report loads, a 30-second admission deadline, 503 plus `Retry-After: 5` when busy, 410 for an expired snapshot, 409 for a snapshot mismatch. No session state, no 202 polling. Telemetry is allow-listed and sanitised - no search term, UPN or filter value reaches it.

## Tests

- **18** SQL semantics tests against a scratch database (dedup of overlapping memberships, unknown vs measured zero, Copilot coverage cases, deterministic tie-breaking, SKU cap, demographic bounding).
- **28** pure API/cache/domain/lifecycle tests (deadline, retained slot, late-generation eviction, telemetry flush, export 410/409).
- **85** portal tests, plus `tsc --noEmit` and `vite build` clean.
- An opt-in release-scale harness that now reports **Inconclusive** rather than Passed when it asserts nothing.

## Review rounds

Four independent models reviewed round 1 (backend correctness, upgrade/operability, UX, security/privacy). Backend, upgrade and security came back clean; UX found real defects, fixed here. Round 2 then found a **blocker in my own round-1 fix** - the search sanitiser called `.trim()` on every keystroke, making the space bar unusable - fixed and pinned with a keystroke-by-keystroke test.

## Reviewer hotspots

- `LicenceActivitySql.cs` (large): the wide-window score path is where the timeout lives.
- `LicenceActivitySnapshotCache.cs`: admission deadline, slot retention, late-generation publication.
- `LicenceActivityAPIController.cs`: role re-check on export and the cache scope derivation.

Addresses #436
Addresses #437

Diff: 65 files changed, 15708 insertions(+), 94 deletions(-)

## Final checkpoint and acceptance status

Feature source head: `f94fb84fe7e45f4efec3b2062cb01ba13db83769`; 3 commits, now merged into `dev` as `b415bc08`. Required CI is green. The user explicitly authorised the admin override for the dev merge; `main` remains untouched. The feature is actually included in **draft release PR #458**, whose performance hold is explicit. The later critique iterations fixed sequential-space typing and C1 control-character handling; those repairs are included at this head.

## Completed Azure SQL performance review: S4 / 200 DTUs and S6 / 400 DTUs

The user authorised a new isolated Azure benchmark environment, followed by mandatory deletion. The same verified synthetic corpus was tested sequentially on the two actual service objectives: 300,000 users, 50 SKUs, 2,060,681 current assignment links and 39 million workload rows. Product `b415bc08` was tested without schema, range, interface or production-timeout changes. Draft optimisation candidate #457 was **not** part of this Azure run.

**Both tiers failed acceptance.** The complete matrix requires 1,500 cold user cases and 1,500 immediate warm partners per tier; each tier completed the 750 narrow-window pairs, then stopped at the first exact 180-day overview returning 503. The measured process-memory growth also exceeded the predeclared 128-MiB limit.

| Measured case | S4 / 200 DTUs | S6 / 400 DTUs |
|---|---:|---:|
| Instrumented isolated 7-day cold HTTP median / max | 11.937 / 12.429 s | 7.341 / 7.910 s |
| Immediate warm overview median / max | 2 / 3 ms | 2 / 3 ms |
| Cold overview logical reads | 1,330,799 | 1,330,799 |
| Exact 180-day HTTP median / max, all 503 | 23.026 / 23.036 s | 21.736 / 21.755 s |
| Narrow cold users median / p95 / max | 2.559 / 5.970 / 8.399 s | 2.039 / 5.284 / 11.014 s |
| Narrow warm users median / p95 / max | 8 / 14 / 80 ms | 8 / 21 / 142 ms |
| Maximum user JSON | 489,852 bytes | 489,852 bytes |
| Four distinct cold loads, max group latency | 12.446 s | 8.250 s |
| Eight duplicate cold callers, max latency | 7.822 s | 6.381 s |
| Excel snapshot max latency | 189 ms | 169 ms |
| Managed / private process growth | 135.4 / 146.6 MiB | 152.4 / 160.5 MiB |
| Completed cold/warm pairs | 750 / 1,500 | 750 / 1,500 |

Warm views and Excel exports issued zero SQL loads/reads. Duplicate cold callers shared one store load; cancellation left the surviving caller successful. All narrow-window deep-page/search/demographic cases completed.

These HTTP measurements use the real attributed API/cache/store/export in a shared .NET Framework client/server test process, including the SQL network path and `STATISTICS IO/TIME`. SQL buffers and plan caches were retained/uncontrolled; this was not a SQL-cold or IIS/App Service/Entra deployment test. Crucially, the first following **uninstrumented** 7-day adapter repeat timed out at both tiers, so the instrumented narrow medians are not proof of reliable uninstrumented performance. No three-successful-repeat uninstrumented adapter median exists.

Actual plans captured wide per-sample hash-aggregation spills downstream of a fast existing columnstore scan, and wide common/tenant Teams queries selected millions of clustered-key lookups plus a sort spill despite an existing covering columnstore path. Query Store also captured plan differences for aborted single-sample Copilot statements. These establish optimisation targets, not that a permanent index or migration is necessary. A separately reviewed fix and new before/after measurements are still required.

All benchmark databases, logical servers, perimeter resources and dedicated resource groups were deleted. Independent Azure API reads confirmed every owned group absent; protected connection/token state was cleared. No customer/production data or existing resources were used.

The full evidence and stable-release hold are summarised in #458. Neither that release PR nor candidate #457 is authorised for an automatic merge.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -175,6 +175,10 @@ jobs:
           -p:ProcessorArchitecture=${{ env.Build_ProcessorArchitecture }}
         env:
           Configuration: ${{ matrix.configuration }}
+      - name: Execute portal tests
+        if: needs.setup_tests.outputs.run_dotnet == 'true'
+        working-directory: src\AnalyticsEngine\Web\Scripts\portal
+        run: npm run test
       - name: Execute unit tests
         if: needs.setup_tests.outputs.run_dotnet == 'true'
         uses: jesusfer/vstest-action@f5ec12ddb71f6e151b738fa4740f8fea61e9214b
--- a/src/AnalyticsEngine/Common/DataUtils/AnalyticsLogger.cs
+++ b/src/AnalyticsEngine/Common/DataUtils/AnalyticsLogger.cs
@@ -377,7 +377,8 @@ namespace DataUtils
             HealthCheck,
             ImporterHeartbeat,
             CopilotAdoptionAnalysis,
-            CopilotAdoptionLifecycle
+            CopilotAdoptionLifecycle,
+            LicenceActivityLifecycle
         }
     }
 }
--- a/src/AnalyticsEngine/Common/Entities/Entities.csproj
+++ b/src/AnalyticsEngine/Common/Entities/Entities.csproj
@@ -81,6 +81,12 @@
   <ItemGroup>
     <Compile Include="BaseEntity.cs" />
     <Compile Include="BuildConstants.cs" />
+    <Compile Include="LicenceActivity\ILicenceActivityStore.cs" />
+    <Compile Include="LicenceActivity\LicenceActivityModels.cs" />
+    <Compile Include="LicenceActivity\LicenceActivityQuery.cs" />
+    <Compile Include="LicenceActivity\LicenceActivityWorkbook.cs" />
+    <Compile Include="LicenceActivity\LicenceActivitySql.cs" />
+    <Compile Include="LicenceActivity\SqlLicenceActivityStore.cs" />
     <Compile Include="Config\AppConfig.cs" />
     <Compile Include="Config\AppConnectionStrings.cs" />
     <Compile Include="CopilotAdoption\AdoptionBand.cs" />
--- a/src/AnalyticsEngine/Common/Entities/LicenceActivity/ILicenceActivityStore.cs
+++ b/src/AnalyticsEngine/Common/Entities/LicenceActivity/ILicenceActivityStore.cs
@@ -1,0 +1,67 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Common.Entities.LicenceActivity
+{
+    public interface ILicenceActivityStore
+    {
+        Task<LicenceActivityOverview> LoadOverviewAsync(
+            LicenceActivityQuery query, LicenceActivitySources sources,
+            ILicenceActivityDiagnostics diagnostics, CancellationToken cancellationToken);
+
+        Task<LicenceActivityUsers> LoadUsersAsync(
+            LicenceActivityOverview overview, LicenceActivityQuery query, LicenceActivitySources sources,
+            ILicenceActivityDiagnostics diagnostics, CancellationToken cancellationToken);
+    }
+
+    public interface ILicenceActivityDiagnostics
+    {
+        void Stage(string stage, long elapsedMs = 0);
+    }
+
+    public sealed class NullLicenceActivityDiagnostics : ILicenceActivityDiagnostics
+    {
+        public static readonly NullLicenceActivityDiagnostics Instance = new NullLicenceActivityDiagnostics();
+        private NullLicenceActivityDiagnostics() { }
+        public void Stage(string stage, long elapsedMs = 0) { }
+    }
+
+    public sealed class LicenceActivitySources
+    {
+        public bool UserMetadata { get; set; }
+        public bool UsageReports { get; set; }
+        public bool CopilotUsageReports { get; set; }
+        public bool CopilotAudit { get; set; }
+        public bool CopilotInteractions { get; set; }
+        public DateTime NowUtc { get; set; }
+
+        public string CacheKey => string.Join(":", UserMetadata, UsageReports, CopilotUsageReports,
+            CopilotAudit, CopilotInteractions, NowUtc.ToString("yyyyMMdd", System.Globalization.CultureInfo.InvariantCulture));
+    }
+
+    public static class LicenceActivityRules
+    {
+        public const string Method =
+            "Activity bands describe the share of observed reporting samples with activity: zero = none, "
+            + "low = below 25%, moderate = 25% to below 75%, high = at least 75%. "
+            + "Incomplete evidence is unknown, not zero. Counts are workload-specific snapshot averages, not daily events.";
+
+        public const string AssignmentCaveat =
+            "Historical activity is shown against currently imported licence assignments, not historical ownership. "
+            + "A user can hold several SKUs; summed assignments are not a unique-user total. Assigned seats are not purchased capacity.";
+
+        public const string InterpretationCaveat =
+            "Activity by a licence holder does not prove that this SKU enabled it. These estimates do not measure productivity, "
+            + "ROI or compliance and are not sufficient evidence to remove a licence.";
+
+        public static string Band(int activeSamples, int observedSamples, int expectedSamples)
+        {
+            if (expectedSamples <= 0 || observedSamples != expectedSamples || activeSamples < 0 || activeSamples > observedSamples)
+                return "unknown";
+            if (activeSamples == 0) return "zero";
+            if ((long)activeSamples * 4 < expectedSamples) return "low";
+            return (long)activeSamples * 4 < (long)expectedSamples * 3 ? "moderate" : "high";
+        }
+    }
+}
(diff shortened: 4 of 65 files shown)


sambetts:
## Verification status at `f94fb84f`

**All three required checks pass** on the current head: `test_dotnet (Release)`, `test_aitracker`, `gitleaks`. Every other build job is green too. The hosted `test_dotnet (Release)` run covers the new portal suite and the new .NET tests, including the SQL semantics tests against a scratch database, so those are not just green on my machine.

Local verification of the exact committed tree (built in isolation so it could not pick up a stale binary):

| Suite | Result |
|---|---|
| SQL semantics (scratch DB) | 18 pass |
| Pure API / cache / domain / lifecycle | 28 pass |
| Release-scale harness | correctly **Skipped** (Inconclusive) when opted out |
| Portal vitest, single worker | 85 pass |
| `tsc --noEmit` / `vite build` | exit 0 / exit 0 |

## Why I have not merged this

Two independent reasons, both of which I think are the right call rather than obstacles to route around:

**1. The measured performance blocker in the description is real and unresolved.** The 180-day preset exceeds the 20-second per-command timeout at 300,000 users in both the B-tree fallback and the production columnstore shape. Fixing it properly needs a query rewrite; the two shortcuts available to me - adding an index or relaxing the timeout - are exactly the two things not approved, and either would hide the problem rather than solve it.

**2. I cannot obtain a legitimate code-owner approval.** `CODEOWNERS` assigns `*` to @sambetts, who is the author of this PR, and GitHub does not permit self-approval. `reviewDecision` is `REVIEW_REQUIRED` with zero reviews. @jesusfer is a code owner for `/.github/` (which this PR touches) and has been requested, but that does not satisfy ownership of the remaining files. Branch protection has `enforce_admins: false`, so an admin merge would technically succeed - I am deliberately not doing that, because it would bypass the review gate rather than satisfy it.

## Review history

Four rounds, seven reviewers, six distinct models, looped until a round came back clean:

- **R1** - backend/SQL correctness (Opus 4.8) **clean**; upgrade/operability (Gemini 3.8) **clean**; security/privacy (Grok 4.6) **clean**, including an explicit trace of the pre-existing telemetry pipeline to confirm the new search term cannot leak through it; UX (Opus 5) found 1 HIGH + 2 MEDIUM.
- **R2** - (GPT-5.6 Sol) found a **blocker in R1's own fix**: the search sanitiser trimmed on every keystroke, making the space bar unusable. A separate evidence audit (Grok 4.6) found my first-draft performance claims overclaimed, and caught that I had measured the *fallback* index shape rather than the production columnstore - which is why the table above has both.
- **R3** - (Opus 4.8) found a **MEDIUM in R2's fix**: the client stripped C0 and DEL, but the server's `char.IsControl` also covers C1, so a pasted U+0085 still produced a 400.
- **R4** - (GPT-5.6 Terra) **clean**, which is what ended the loop.

Every finding was reproduced against the code before being acted on, and every fix is pinned by tests in **both** directions - the guard fires when it should and stays silent when it should not.

## Suggested next step

Land the query rewrite for the wide window first, re-run the release-scale harness in columnstore mode, and only then take this through review. The harness is already opt-in via `LICENCE_ACTIVITY_PERF=1` and reuses a retained database, so re-measuring is cheap.

PR #472, opened 2026-09-08
Measure licence activity by imported week, not by one pinned Sunday

sambetts:
## Problem

A production tenant reported the Licence activity page as over 97% **Unknown** for every Microsoft 365 service, while the measured minority came back as **100% High** with literally zero Moderate/Low/No-activity. That shape is the diagnosis: it only happens if presence in the data already implies activity.

Three separate things combined.

1. `dbo.teams_user_activity_log` and its siblings are Graph's **daily** user-detail reports, fetched as `...(date=YYYY-MM-DD)` in `AbstractDailyActivityLoader.LoadReportPageForDateFromGraph`. Those return **only the people who did something on that day**. The repo already documents this in `CopilotAdoptionSqlIntegrationTests.cs:390` — *"a snapshot landing on a Saturday emptied the tab entirely"*.
2. Licence activity took **one reading per week**: `TOP 1 ... ORDER BY date DESC` inside the week (`LicenceActivitySql.AppendM365Overview`). `#Weeks` is anchored to `1900-01-01`, a **Monday**, so `m365_to` — and therefore the reading — always landed on the week's **Sunday**.
3. A person was banded only when present in **every** reading: `scores.frequency_known = 1 AND scores.observed_samples = coverage.expected_samples`.

Net effect: *"measured"* meant *"used the service on every Sunday of the period"*. Everyone else fell into Unknown, which the UI correctly refuses to render as zero — so the page was honest but nearly empty.

## Change

A reading becomes a **week that was imported day by day**, and absence across such a week becomes evidence of no activity rather than absence of evidence.

### `LicenceActivitySql.cs`
- New `#Days` spine in the overview preamble: every calendar day each week covers.
- `#Samples` now takes one row per week (keyed by the week's last day) **only when every day of that week has a report** and the week ends on or before `@settled`. Implemented as an `OUTER APPLY ... TOP (1)` per day — at most 180 single-row index seeks, bounded by days in the period rather than rows in the table.
- `@complete{n}` is gone; status is decided purely on covered-week count.
- `AppendM365OverviewScore` rewritten: scores across the whole week, counts active weeks, then inserts an explicit **measured zero** for every eligible person with no activity.
- `AppendM365Users` and `AppendM365ReadModel` were two divergent implementations of the same scoring. They are now **one**; `AppendM365ReadModel` and the `M365OverviewScores` constant are deleted.

### `LicenceActivityReadModel.cs`
- `Evidence()` resolves an absent person to a complete, zero reading instead of `missingCoverage`/Unknown.
- A person whose **own** reading count falls short still reports `partial` and is not banded. Absence is now meaningful; *incomplete* evidence still is not.
- The read model deliberately does **not** materialise the inactive majority — at 200k users that would be one cached dictionary entry per person per workload. It applies the rule in memory instead.
- `IsOfficialReport` replaced by an explicit `presenceGated` rule.

### Group-filter guard (the reason this isn't a one-line change)
`UserGroupsFilter` scopes the **usage-report** import (`AbstractUserDailyActivityLoader.IdInScope`) but **not** the **user** import — `ProductionGraphImportSectionFactory` builds `UserMetadataUpdater` without it. So the two populations differ, and "absent ⇒ inactive" would be a confident wrong answer about people nobody ever looked at.

When a group filter is configured the M365 workloads keep the old presence gate and the report explains why (`Notes.UsageReportsGroupFiltered`). Enforced in both paths: `@groupFiltered = 0` in SQL, `presenceGated` in the read model. New `LicenceActivitySources.UsageReportsGroupFiltered`, included in `CacheKey` and copied by `CachedLicenceActivityStore.Capture` so a cached snapshot can't leak across the two modes.

## Performance — a real regression was found here, and fixed

Covering every day instead of one Sunday in seven is ~7x more data. The first implementation cost far more than 7x, because of a **plan** problem rather than a volume one.

Both scorers joined the large usage table to the tiny `#Samples`/`#Weeks` temp tables by a **date range**:

```sql
JOIN #Samples AS chosen ON chosen.workload = N
 AND activity.[date] >= chosen.m365_from
 AND activity.[date] < chosen.end_exclusive
```

A non-equi join gives the optimiser no equality to hash on. The captured plan showed
`Nested Loops[[teams_user_activity_log]][[NCCI_teams_user_activity_log_metrics]]` — a nested loop
driving into the **nonclustered columnstore**, which cannot seek. Every covered week therefore
re-read the whole activity table.

Commit `7c8a3b25` matches the week by **equality** on a week start derived arithmetically from each
report row's own date, and collapses the per-user aggregate to two steps so
`COUNT(DISTINCT sample_date)` (which forced a spool that replayed the ~8.4M-row day-level stream)
becomes a plain `SUM`.

The join key is the week's true, **unclamped** Monday. `sample_date`, `m365_from` and `m365_to` are
all clamped to the requested period, so none of them can be derived from a report row's own date;
`week_start` can. The surrounding `@from`/`@endExclusive` predicate reproduces the clamping exactly,
so the row set is unchanged:
`date ∈ [week_start, week_start+6] ∧ date ∈ [@from, @to]  ≡  date ∈ [m365_from, end_exclusive)`.

### Measured

Synthetic 300k-user / 50-SKU harness: 8.44M rows per usage table, **180 imported days**, columnstore
mode (production ships the NCCI via `202608310800001_ColumnstoreUsageReportMetrics`; B-tree is only
the documented fallback). Measured on `LoadReadModelAsync` — the path the portal actually uses.

| Metric (Teams workload) | Before | After | Change |
|---|---|---|---|
| Logical reads | 35,024,453 | **447,751** | **78x fewer** |
| Read-model load, 7-day window | 34,540 ms | **16,519 ms** | 2.1x faster |
| Read-model load, 180-day window | 420,593 ms | **238,988 ms** | 1.8x faster |

Logical reads for the other three workloads moved the same way (Outlook 27.5M → 437,606;
OneDrive 26.8M → 417,360; SharePoint 26.5M → 417,220).

### Plan operators — why it got faster

Captured with `SET STATISTICS XML`, filtered to operators touching `teams_user_activity_log`:

| | Operators on `NCCI_teams_user_activity_log_metrics` |
|---|---|
| Before | `Index Scan`, **`Hash Match`**, **`Nested Loops`**, `Compute Scalar` |
| After | `Index Scan`, `Filter`, `Compute Scalar` |

The **`Nested Loops` driving into the nonclustered columnstore is gone**. A columnstore has no seek,
so each loop iteration re-scanned it; the scorer now makes a single pass. (The `Nested Loops` + `Top`
+ `Index Seek` that remain on the narrow `IX_date` index are the *coverage probe* — the
`OUTER APPLY ... TOP (1)` per day in `AppendM365Overview`, which is by design at most 180 single-row
seeks and is not the scorer.)

Before the fix the **7-day** window failed the HTTP acceptance matrix outright with
`ServiceUnavailable` ("Licence activity could not be loaded. Retry the request."). It now completes
in 6.7–8.0 s per workload.

**Environment caveat, stated plainly:** this is LocalDB on a shared, heavily contended developer box
(8 logical cores, ~92–100% CPU from other processes throughout). LocalDB is Express-class: **no
intra-query parallelism**, and the four M365 workloads run *concurrently*, so they serialise onto one
core. These elapsed figures are **diagnostic only and are not Azure SQL 200–400 DTU acceptance
evidence** — the harness's own docstring says so. Logical reads are the load-independent signal, and
that is the axis with the 78x improvement.

### Known remaining risk — reviewer decision needed

At the **180-day** window the four M365 workloads still hit the 20 s
`LicenceActivitySql.CommandTimeoutSeconds` on this harness (all four land at ~20.3 s together, which
is the signature of them serialising on a single core). The 7-day window passes comfortably.

Scope of that risk: `LicenceActivityQuery.MaximumDays` is 180, so 180 days is the **worst case an
admin can select**, not the normal one — the **default window is 28 days**
(`start = end.AddDays(-27)`). So the default page load sits near the measured-good end, and the
unproven case is an explicit opt-in to the maximum range.

I could not settle whether the 180-day timeout is a genuine production limit or an artefact of a
single-core, 4-way-concurrent, 100%-CPU box. A `dev` baseline **cannot** answer it either: on daily,
positive-only data `dev` bands almost nobody, so `LicenceActivityLoadTests` aborts with
*"do not benchmark an all-unknown shortcut"* — `dev` is fast because it is broken, not because it is
efficient (2,249 logical reads vs 447,751 for the same 7-day Teams query). Please treat the widest
reporting window as unproven at 300k users until it is measured on real Azure SQL.

## Migrations / config

None. No `Migrations/` change, no `Create DB.sql` change, no `CONFIG_VERSION` change. The new
`week_start` column is on the `#Samples` **temp** table only. `LicenceActivitySources.UsageReportsGroupFiltered`
is derived at request time from the existing `UserGroupsFilter` app setting; it is not persisted installer config.

Because there is no schema migration, the repo's "prove every schema change" gate does not formally
bind — but the before/after above is recorded here and in the code comments so the decision can be
re-checked.

## Reviewer hotspots

- `AppendM365Users` / `AppendM365OverviewScore` — the equality-join rewrite and its clamping argument.
- `LicenceActivityReadModel.Evidence` — the three-way split between presence-gated, measured-zero and partial.
- `CachedLicenceActivityStore.Capture` — a missed field here would silently serve the wrong mode from cache.
- `average_actions` is deliberately kept a **per-day** figure so the number an admin already knows doesn't silently become a weekly total. The rewrite preserves it exactly: the action expressions are `CAST(ISNULL(col,0) AS float) + ...`, so they are non-NULL floats and `AVG(x) ≡ SUM(x)/COUNT(*)`.

## Tests

`Tests.UnitTests` — **110 licence-activity tests pass**, and `test_dotnet (Release)` is green.

The scale/integration fixture seeded **one Sunday per week**, i.e. it modelled an import that only ever ran on Sundays — which this change (correctly) reports as unmeasured. `SeedOneUsageTable` now seeds **every day** of each week, and the 300k-user scale seed is now **daily and positive-only** (rows only for people active that day), matching what a healthy daily import leaves behind. `RetainedMarker` bumped so a stale retained scale DB can't be reused.

Renamed/added:
- `PerUserMissingRows_AreUnknownAndNeverLeastActive` → `PerUserMissingRows_AreMeasuredZeroWhileMissingReportDaysStayUnknown`
- **new** `AMissingReportDay_MakesItsWholeWeekUnknownForEveryone`
- **new** `AGroupFilteredUsageImport_LeavesUnmeasuredPeopleUnknownRatherThanInactive`
- `DistinctPinnedSampleCounts_...` → `DistinctWeeklyReadings_PreserveZeroDuplicatesAndExpectedBoundary`

Portal: `npm run build` (`tsc --noEmit` + `vite build`) clean. `bands.ts` copy updated to describe weeks.

## User-visible effect

Most people move from **Unknown** to a real band, and "No activity" becomes a real, common answer instead of a near-impossible one. Deployments that scope the import with `UserGroupsFilter` keep the old conservative behaviour plus an explanatory note.

Addresses the "everything is Unknown" report.

### DIFF
--- a/src/AnalyticsEngine/Common/Entities/LicenceActivity/CachedLicenceActivityStore.cs
+++ b/src/AnalyticsEngine/Common/Entities/LicenceActivity/CachedLicenceActivityStore.cs
@@ -81,6 +81,7 @@ namespace Common.Entities.LicenceActivity
                 CopilotUsageReports = sources.CopilotUsageReports,
                 CopilotAudit = sources.CopilotAudit,
                 CopilotInteractions = sources.CopilotInteractions,
+                UsageReportsGroupFiltered = sources.UsageReportsGroupFiltered,
                 NowUtc = sources.NowUtc
             };
     }
--- a/src/AnalyticsEngine/Common/Entities/LicenceActivity/ILicenceActivityStore.cs
+++ b/src/AnalyticsEngine/Common/Entities/LicenceActivity/ILicenceActivityStore.cs
@@ -34,19 +34,33 @@ namespace Common.Entities.LicenceActivity
         public bool CopilotUsageReports { get; set; }
         public bool CopilotAudit { get; set; }
         public bool CopilotInteractions { get; set; }
+
+        /// <summary>
+        /// True when <c>UserGroupsFilter</c> scopes the Microsoft 365 usage-report import to particular
+        /// Entra groups.
+        ///
+        /// This matters because the USER import is not filtered while the usage-report import is, so the
+        /// two populations differ. Normally a person with no rows across a fully imported week is proof
+        /// that they did nothing; under a group filter they may simply never have been looked at, and
+        /// reporting them as "No activity" would be a confident wrong answer. When this is set the
+        /// report falls back to leaving those people Unknown.
+        /// </summary>
+        public bool UsageReportsGroupFiltered { get; set; }
+
         public DateTime NowUtc { get; set; }
 
         public string CacheKey => string.Join(":", UserMetadata, UsageReports, CopilotUsageReports,
-            CopilotAudit, CopilotInteractions, NowUtc.ToString("yyyyMMdd", System.Globalization.CultureInfo.InvariantCulture));
+            CopilotAudit, CopilotInteractions, UsageReportsGroupFiltered,
+            NowUtc.ToString("yyyyMMdd", System.Globalization.CultureInfo.InvariantCulture));
     }
 
     public static class LicenceActivityRules
     {
         public const string Method =
-            "Activity levels describe how much of the measured period someone was active in: "
-            + "No activity = never, Low = under a quarter, Moderate = a quarter to under three quarters, "
-            + "High = three quarters or more. Where the period could not be measured in full the level is "
-            + "Unknown, not zero.";
+            "Activity levels describe how many of the period's weeks someone was active in: "
+            + "No activity = none, Low = under a quarter, Moderate = a quarter to under three quarters, "
+            + "High = three quarters or more. A week is only counted when every one of its days was "
+            + "imported; where a week could not be measured in full the level is Unknown, not zero.";
 
         public const string AssignmentCaveat =
             "Past activity is shown against who holds each licence today, not who held it at the time. "
@@ -126,6 +140,11 @@ namespace Common.Entities.LicenceActivity
 
             public const string DemographicsCapped =
                 "The department and country breakdowns show only the 50 largest of each.";
+
+            public const string UsageReportsGroupFiltered =
+                "This deployment only collects Microsoft 365 usage for people in particular Entra groups, "
+                + "but it lists everyone who holds a licence. Anyone outside those groups is shown as "
+                + "Unknown rather than as doing nothing, because they were never measured.";
 
             public const string RankingMethod =
                 "The most and least active lists rank people by how often they were active in the chosen service, "
--- a/src/AnalyticsEngine/Common/Entities/LicenceActivity/LicenceActivityReadModel.cs
+++ b/src/AnalyticsEngine/Common/Entities/LicenceActivity/LicenceActivityReadModel.cs
@@ -70,6 +70,7 @@ namespace Common.Entities.LicenceActivity
         private readonly IReadOnlyDictionary<int, LicenceActivityScore>[] _scores;
         private readonly sbyte[][] _bands;
         private readonly RankValue[][] _rankValues;
+        private readonly bool _usageReportsGroupFiltered;
         private readonly object _rankingGate = new object();
         private readonly ConcurrentDictionary<string, RankingIndex> _rankings =
             new ConcurrentDictionary<string, RankingIndex>(StringComparer.Ordinal);
@@ -80,7 +81,8 @@ namespace Common.Entities.LicenceActivity
             IReadOnlyList<LicenceActivityDirectoryUser> users,
             IReadOnlyList<LicenceActivityMembership> memberships,
             IReadOnlyList<LicenceActivityCoverage> coverage,
-            IReadOnlyDictionary<string, IReadOnlyDictionary<int, LicenceActivityScore>> scores)
+            IReadOnlyDictionary<string, IReadOnlyDictionary<int, LicenceActivityScore>> scores,
+            bool usageReportsGroupFiltered = false)
         {
             if (range == null) throw new ArgumentNullException(nameof(range));
             if (licences == null) throw new ArgumentNullException(nameof(licences));
@@ -91,6 +93,7 @@ namespace Common.Entities.LicenceActivity
             if (range.DepartmentId.HasValue || range.CountryId.HasValue || range.LicenceTypeId.HasValue)
                 throw new ArgumentException("A licence activity read model must cover an unfiltered date range.", nameof(range));
 
+            _usageReportsGroupFiltered = usageReportsGroupFiltered;
             _from = range.From;
             _to = range.To;
             Id = Guid.NewGuid().ToString("N");
@@ -266,6 +269,8 @@ namespace Common.Entities.LicenceActivity
             else if (overview.DistinctAssignedUsers == 0)
                 overview.Messages.Add(LicenceActivityRules.Notes.NobodyHoldsALicence);
             overview.Messages.Add(LicenceActivityRules.Notes.NoDisplayNames);
+            if (_usageReportsGroupFiltered)
+                overview.Messages.Add(LicenceActivityRules.Notes.UsageReportsGroupFiltered);
             foreach (var item in overview.Coverage.Where(item => item.Status != LicenceActivitySql.Available))
             {
                 if (!string.IsNullOrWhiteSpace(item.Message))
@@ -549,7 +554,16 @@ namespace Common.Entities.LicenceActivity
             var status = coverage.Status ?? LicenceActivitySql.MissingCoverage;
             var active = present ? score.ActiveSamples : 0;
             var observed = coverage.ObservedSamples;
-            if (IsOfficialReport(coverage.Source))
+
+            // Sources where being absent proves nothing, so only the people we actually have rows for
+            // can be banded at all:
+            //  - the Copilot usage report, a rolling per-person report that only ever lists people who
+            //    hold a Copilot licence, so absence means "not licensed" rather than "did nothing";
+            //  - the Microsoft 365 usage reports WHEN a group filter scopes their import, because the
+            //    user import is not filtered and absence then means "never looked at".
+            var presenceGated = coverage.Source == LicenceActivitySql.CopilotReportSource
+                || (coverage.Source == LicenceActivitySql.M365ReportSource && _usageReportsGroupFiltered);
+            if (presenceGated)
             {
                 observed = present ? score.ObservedSamples : 0;
                 if (status == LicenceActivitySql.Available)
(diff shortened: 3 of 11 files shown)


sambetts:
## Performance measurement — this is NOT ready to merge

Measured tonight on the synthetic 300k-user / 50-SKU harness, **columnstore index mode**, with `dev` and this branch run against the **same retained database** so both versions read identical data. LocalDB on a developer box: diagnostic only, not Azure SQL 200–400 DTU acceptance evidence, but the two sides are directly comparable.

The scale seed had to change to measure this at all: it previously seeded **one Sunday per week**, which models an import that only ever ran on Sundays. It now seeds **every day, positive-only** — 8.44M rows per workload table across 180 distinct report days. `dev` was then run against that same daily data, so this is a like-for-like comparison, not old-shape vs new-shape.

### Overview path (`SqlStore_AtReleaseScale`, teams workload)

| Window | `dev` reads | `dev` elapsed | This branch reads | This branch elapsed |
|---|---|---|---|---|
| 7 days | 7,329 | 4.4 s | 161,594 | 7.8 s |
| **180 days** | 18,937 | **7.9 s** | 84,144 | **141.6 s** |

Then, replacing the `COUNT(DISTINCT ...)` over the raw join with a two-step aggregate (collapse to one row per (user, week), then sum):

| Window | This branch, two-step aggregate |
|---|---|
| 7 days | 161,484 reads / 9.1 s |
| **180 days** | 823,919 reads / **67.4 s** |

So the wide window improved 141.6 s → 67.4 s, but **logical reads rose to 824k** and it is still **~8.5× slower than `dev`**. Elapsed and reads moving in opposite directions is the signature this repo already documents in *"Logical reads alone can choose the slower index"* — this is a plan problem, not just volume.

### What this means

The functional change is right and its tests pass (110 licence-activity tests, including new coverage for a missing report day and for the group-filter guard). But reading every day of every week, instead of one pinned day, costs far more than expected on the overview path, and I have not found a shape that pays for itself. **Not approved for merge in this state.**

Two notes that bound the blast radius:

- `SqlLicenceActivityStore.LoadOverviewAsync` — the path measured above — is **not** on the production request path. The controller wraps the store in `CachedLicenceActivityStore`, which only calls `LoadReadModelAsync`. An earlier run of the cached path completed each workload in 6.8–8.7 s at 300k users, but its HTTP acceptance matrix still failed, so the production path is **also unproven**.
- Both `dev` **and** this branch exceed the harness's 20 s command timeout on the 7-day acceptance scenario against this daily-seeded database (`dev` 25.2 s, branch 28.8 s). That specific failure pre-dates this change and is partly a property of the box.

### Next steps

1. Profile the wide-window overview plan with `LICENCE_ACTIVITY_PERF_DIAGNOSTIC_SHOWPLAN=1` and find why 824k reads are needed to touch ~8.4M rows once.
2. Consider bounding the join differently — the weeks tile the period contiguously, so the week key could be derived arithmetically instead of joined as a range against `#Weeks`.
3. Re-measure the **cached read-model** path (the one production uses) before/after on the same retained DB; that is the number that actually decides shippability.
4. Re-run the acceptance matrix once the wide window is inside budget.

### Reusable harness notes

- Fixed a latent bug: the retained-scale marker was **written as a hard-coded literal** but **validated against the `RetainedMarker` constant**, so a retained database could never be reused after a version bump. Retained reuse now works — a run costs ~30 s instead of a ~25 minute reseed.
- Repro: set `LICENCE_ACTIVITY_PERF=1`, `LICENCE_ACTIVITY_PERF_RETAIN_DB=1`, `LICENCE_ACTIVITY_PERF_INDEX_MODE=columnstore`, `LICENCE_ACTIVITY_PERF_STATE_FILE=<absolute out-of-band path>`, `LICENCE_ACTIVITY_PERF_DIAGNOSTIC=1`, `LICENCE_ACTIVITY_PERF_DIAGNOSTIC_TIMEOUT=300`, `LICENCE_ACTIVITY_PERF_DIAGNOSTIC_WINDOW=narrow|wide`.

sambetts:
## Session update — performance regression found and fixed; one open decision

Picking this up from the draft state, where the body said the performance question was unsettled. It is now settled in one direction and explicitly **not** in the other.

### What was wrong

The weekly rewrite joined the large usage tables to `#Samples` / `#Weeks` by a **date range**. A non-equi join gives the optimiser no equality to hash on, so it chose a `Nested Loops` driving into the nonclustered **columnstore** — which cannot seek — and re-read the activity table once per covered week.

On the 300k-user / 8.44M-row-per-table harness this made the page fail outright: the **7-day** window took 34.5 s against the hard 20 s `CommandTimeoutSeconds` and returned `ServiceUnavailable`. The previous body's claim of "6.8–8.7 s per workload, inside the 20 s command timeout" was wrong.

Note the `dev` baseline **cannot** be used for an A/B here: on daily, positive-only data `dev` bands almost nobody, so `LicenceActivityLoadTests` aborts with *"do not benchmark an all-unknown shortcut"*. `dev` is fast because it is broken (2,249 logical reads vs 447,751 for the same 7-day Teams query), not because it is efficient.

### Fix (`7c8a3b25`)

Match the week by **equality** on a week start derived from each report row's own date, and collapse the per-user aggregate to two steps so `COUNT(DISTINCT ...)` stops forcing a spool.

| Teams workload | Before | After |
|---|---|---|
| Logical reads | 35,024,453 | **447,751** (78x fewer) |
| 7-day read-model load | 34,540 ms | **16,519 ms** |
| 180-day read-model load | 420,593 ms | **238,988 ms** |

Plan operators on `NCCI_teams_user_activity_log_metrics`: `Index Scan` + **`Hash Match`** + **`Nested Loops`** → `Index Scan` + `Filter`. The nested loop into the columnstore is gone — that is the causal evidence, not just a faster number.

### Verification

- 110 licence-activity tests pass, 0 failed.
- Two independent model reviews of the SQL (gpt-5.6-sol, gemini-3.8-flash) returned **no defects**. The second worked through the equivalence proof explicitly for the first/last partial weeks, single-day periods, `datetime`-vs-`date` typing, `SET DATEFIRST` independence, `SUM` vs `COUNT(DISTINCT)`, `SUM/SUM` vs `AVG`, other `#Samples` consumers, and the fully-imported-week gate.

### The open decision — I have deliberately not merged this

At the **180-day** window the four M365 workloads still hit the 20 s command timeout on the harness, all landing at ~20.3 s together. That is the signature of them serialising: LocalDB is Express-class with **no intra-query parallelism**, the four workloads run concurrently, and the box was at 92–100% CPU from other processes throughout. So it may well be an artefact — but I cannot prove that here, and this harness is explicitly documented as *not* Azure SQL acceptance evidence.

Scope: `LicenceActivityQuery.MaximumDays` is 180, so that is the **worst case an admin can select**; the **default window is 28 days**. The default load sits near the measured-good end.

**This needs either a measurement on real Azure SQL or an explicit accept-the-risk decision.** Merging past it on bypass privileges did not seem like my call to make.

PR #477, opened 2026-09-09
DLP policies impacting Copilot and agents: import, schema and report

sambetts:
## Summary

Adds a single view of **which Copilot actions and agents are being blocked by Microsoft Purview DLP policies, how often, and who is most affected** — new installer toggle, new tables, new import path, and a new portal report page.

## The finding that shaped the design

The task assumed the DLP audit source (`DLP.All`, separate permission) would carry the Copilot blocks. It does not, and building it that way would have shipped a report that could never answer the question.

Verified against [Microsoft Learn — Audit logs for Copilot and AI applications](https://learn.microsoft.com/en-us/purview/audit-copilot) and the [Management Activity API schema](https://learn.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#dlp-schema):

- DLP policies scoped to the **"Microsoft 365 Copilot and Copilot Chat"** location do **not** emit standalone `DlpRuleMatch` records. Microsoft documents the DLP feed's workloads as Exchange Online, Endpoint and SharePoint/OneDrive only — Copilot is absent.
- DLP records carry the literal `UserKey = "DlpAgent"` and **no agent identity anywhere in the schema**.
- The Copilot block signal is embedded in the `CopilotInteraction` record instead, in `AccessedResources[].Status` and `AccessedResources[].PolicyDetails`, **alongside `AgentId` / `AgentName`**.

So the Copilot-embedded path is the *only* source that can attribute a DLP block to an agent — and it needs **no new permission**, because it rides on the Copilot audit import we already have.

The feature therefore has two independent sources, kept visually and structurally separate. They are deliberately **never joined on user + time** to guess an agent; that would manufacture attribution the audit data cannot support.

| Source | Table | Needs `ActivityFeed.ReadDlp`? | Can name the agent? |
|---|---|---|---|
| Copilot interaction records (`Audit.General`) | `copilot_dlp_events` | No — uses the existing `Copilot` toggle | **Yes** |
| `DLP.All` feed | `dlp_rule_matches` | Yes — new `ImportDlp` toggle | No |

### Import
- `AccessedResource` gains `Status`, `PolicyDetails` (`PolicyId` / `PolicyName` / `Rules[]` with `RuleId`, `RuleName`, `Actions[]`, `Severity`, `RuleMode`) and `XPIADetected`; `CopilotEventData` gains `DLPEvaluationDeferred` / `…Reason`. `CopilotAuditEvent` shares the same `AccessedResource` type, so both parallel models see the new fields (pinned by a test).
- **`CopilotDlpRules`** — pure, dependency-free classifier shared by *both* sources, so "blocked" means the same thing everywhere. Conservative in both directions:
  - a rule listing `BlockAccess` while in `RuleMode = "Audit only"` is **not** a block (it is simulation);
  - a bare `Status = "failure"` with no policy detail is **not** DLP at all (deleted file, permissions, …).
  - An unrecognised action is treated as non-blocking — under-report rather than invent blocks.
- `is_blocked` is decided in C# and carried into staging. The merge SQL contains **no** classification logic, so there is exactly one implementation of "what counts as a block" and it is unit-tested.
- New `DlpAuditEventManager` + `IDlpStagingWriter` port (mirrors the Power Platform pattern), two merge scripts, two staging tables.

### Routing — the trap this had to avoid
A DLP record names the workload where the match was **detected** (`SharePoint`, `Exchange`, `Endpoint`), so routing by `Workload` would hand it to the SharePoint/Exchange deserialiser and **silently drop every policy field**. `WorkloadOnlyAuditLogContent` gains `RecordType`, and DLP record types (11, 13, 33, 63, 107) are matched **before** any workload route. Covered by tests in both directions, including that a `RecordType`-less SharePoint event is unaffected.

### Handling the missing permission
`DLP.All` needs `ActivityFeed.ReadDlp`, a **separate** grant from `ActivityFeed.Read` with its own admin consent. Previously `CreateInactiveSubcriptions` rethrew on any subscription failure, so ticking the DLP box before consent would have **taken SharePoint, Copilot and Power Platform importing down with it**. `DLP.All` is now the one *optional* content type: its failure is logged with a specific, actionable message and dropped for the cycle. Mandatory feeds still fail loudly.

### Schema — classification: **purely additive**
Migration `202609090601123_DlpCopilotImpact` creates five new, initially-empty tables (`dlp_policies`, `dlp_rules`, `dlp_actions`, `copilot_dlp_events`, `dlp_rule_matches`) plus FK indexes. Nothing existing is altered, narrowed, dropped or re-indexed, so **no benchmark is required** under the prove-it rule and there is no "before" query to measure. Upgrade time is effectively instant at any tenant size; no maintenance window, importer can keep running.

`copilot_dlp_events` is deliberately a **separate sparse table** rather than extra columns on `copilot_event_accessed_resources`: that is the largest Copilot table and its de-dup merge seeks on the benchmarked `IX_copilot_event_accessed_resources_dedup` (see `WidenCopilotAccessedResourceDedupIndex`). Widening its de-dup tuple would invalidate that measurement and slow the hot import path for every customer, to carry columns that are NULL on almost every row.

`Create DB.sql` is unchanged — like the Copilot and Power Platform tables, these are created only by migrations.

### Config / installer
- New `[ImportProp] ImportDlp`; `CONFIG_VERSION` bumped **2.2.0 → 2.3.0** (minor, additive) with a history line.
- Installer checkbox, Test Configuration coverage entry (declared *unverifiable*, with the reason: the Activity API check reads the subscription list, which needs only `ActivityFeed.Read`, so it cannot prove the DLP grant), and a Health-page label.

### Web
- `DlpAPIController` (`api/Dlp/availability`, `api/Dlp/summary`). All aggregates computed in SQL; Copilot queries filter on the denormalised `copilot_chats.time_stamp` rather than joining `audit_events`.
- New Insights page **"DLP impact"**. Blocked and *Audited only* are always separate columns — merging them would tell an admin their users are blocked when a policy is merely in simulation. Route is always registered; the empty/disabled state is driven by the availability endpoint.

## Verification

- **Full solution builds**; portal `tsc --noEmit && vite build` clean.
- **553 tests run, all passing** (DLP, Copilot, migration, workload-toggle, install-toggle, health, activity-importer). Includes `MigrationSnapshotTests` and `UrlFullUrlMigrationPipelineTests.LatestId` (updated).
- **31 new tests** covering field parsing (with a Greek filename), the block/audit classification in both directions, `RecordType` routing, the optional-subscription rule, and staging.
- **Merge SQL executed for real**, not just parsed:
  - Copilot merge — 3 staged rows → **2 persisted** (the row whose `copilot_chats` parent was absent was dropped by the `INNER JOIN` rather than failing the batch on the FK); **still 2 after a second run** (idempotent); 1 blocked / 1 audited as classified; all six FK dimensions resolved; the Greek resource name joined correctly.
  - `DLP.All` merge — dimension upserts and the NULL-safe `INTERSECT` de-dup verified the same way.
- **Manual upgrade script validated end to end** against a database at the exact prior migration: applies, **resumes correctly after a partial run**, stamps `__MigrationHistory`, and is a **no-op on re-run**. Its stamp carries **this migration's own model blob** (embedded as base64 → `varbinary`), not the predecessor's — copying the predecessor would make EF auto-migrate and fail with *"There is already an object named 'copilot_dlp_events'"*. The pre-stamp guard checks **schema only**, never data state.
- Confirmed by reading `SaveMetadataAsync` that `audit_events` rows are committed by the staging merge *before* `ProcessExtendedProperties` runs, so the merges' `INNER JOIN` to their parent is safe.

## Reviewer hotspots

1. **`CopilotDlpRules.Classify`** — this is the whole meaning of the report. The `RuleMode == "Audit only"` and "failure with no policy detail" exclusions are the two judgement calls.
2. **Dispatcher ordering** — the DLP `RecordType` check must stay above the workload routes.
3. **`AccessedResources[].PolicyDetails` shape is not contractual.** Microsoft describes it only in prose; the published OData schema for `CopilotInteraction` omits it entirely. Everything is optional and nothing is required to deserialise, so an unexpected payload degrades to "no policy detail" rather than dropping the interaction. Worth a second opinion on whether that is defensive enough.

## ⚠️ Merge-order action required

An in-flight sibling branch carries migration **`202609090519337_AgentCostTables`** (commit `546e0871`, not yet on `dev`, no open PR at time of writing). Its id sorts **before** this PR's `202609090601123`.

If `AgentCostTables` merges to `dev` first, this migration becomes the newest in the chain while carrying a snapshot that knows nothing about the cost tables — `MigrationSnapshotTests` will fail and EF would try to auto-migrate them back. **On rebase, re-scaffold this migration's `.resx` snapshot** (`Add-Migration -Force DlpCopilotImpact`, or the headless `MigrationScaffolder` route) and keep the existing `Up()`/`Down()`. It is a couple of minutes' work but it is not optional.

## Deferred

- Sensitive-information-type detail (`ConditionsMatched.SensitiveInformation[]`) is not imported. It needs a further permission (*"Read DLP policy events including sensitive details"*) that this product deliberately does not request — we report that a policy fired, never the content that triggered it.
- `DLPEvaluationDeferred` is counted and logged per cycle (an unknown DLP verdict, neither block nor all-clear) but not persisted; storing it would mean a column on the largest Copilot table for a signal that is normally zero.
- Wiki page for the new report and the `ActivityFeed.ReadDlp` consent step.


sambetts:
Added demo/fake-data coverage for this feature (commit 25a24a9f), so the new page isn't empty in a demo or perf environment:

- **New FakeDataGen menu item** - "Generate fake DLP policy activity (Copilot + tenant-wide)". It attaches policy matches to Copilot interactions that **already exist** rather than inventing its own, because this report is per-agent - invented interactions would make the DLP page and the Copilot pages disagree. Re-running tops the data up instead of doubling every count.
- **Synthetic demo database** now seeds the DLP dimensions and attaches events to roughly 1 in 12 interactions (weighted towards agent traffic), plus the matching tenant-wide rule matches. `ImportDlp=True` added to `DemoPortalReadiness.RequiredImportJobSettings`.

One seeded policy is deliberately left in **"Audit only"** mode: the report's central distinction is between a policy that denied content and one that merely matched, so demo data containing only blocking policies would never demonstrate - or regression-test - it. A new test asserts both states are present, that events stay a small minority of interactions, and that an `Audit only` rule is never recorded as a block.

Full suite after this change: **595 passed, 1 skipped, 0 failed**.

sambetts:
### Fixed: page crashed on load

Reported from a local run of `#/insights/dlp`:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'map')
```

**Root cause was a JSON contract mismatch, not a UI bug.** This web application configures **no** camelCase contract resolver - every API model names its own wire fields explicitly with `[JsonProperty]` (see `SystemStatusApiModel`). `DlpModels.cs` carried none, so it serialised in PascalCase (`Reasons`, `TopAgents`, ...), every field `types/dlp.ts` read came back `undefined`, and the first `.map()` took the page down.

What makes this worth a guard rather than a one-line fix: **the defect was invisible from C#.** The controller, the SQL and the models were each correct in isolation, every existing test passed, and only loading the page in a browser would have caught it.

So the contract is now asserted instead of assumed:

- `DlpModels.cs` declares the wire name of every property.
- **`DlpApiContractTests`** fails the build if a property has no `[JsonProperty]`, if a name isn't camelCase, if a collection the page maps over would serialise as anything but an array, or if `types/dlp.ts` stops declaring a field the API sends - so both halves of the contract are checked together.
- **`DlpPage.test.tsx`** reproduces the reported `TypeError` and pins the degraded behaviour ("Nothing in this period." rather than a white screen). Verified in both directions: reverting the guard makes it fail with the exact reported error.

Also fixed a genuine consequence TypeScript caught: replacing `availability?.reasons.map` with `(availability?.reasons ?? []).map` removed the null-narrowing inside the callback, so `availability.available` became `availability?.available`.

Full solution builds; portal suite **117 passed** (the 4 licence-activity failures seen in a full-suite run are load-related timeouts - they pass in isolation and touch no DLP code).

sambetts:
**Post-merge note: this PR's migration was renumbered and its snapshot regenerated during integration into `dev`.** Recording it here because the merged PR body still describes the original id, which no longer exists.

### What changed and why

This PR and #478 (agent cost) each added migrations on separate branches, and their ids interleaved:

```
202609090519337_AgentCostTables        (#478)
202609090601123_DlpCopilotImpact       (this PR)  <- landed between the other two
202609090647151_AgentCostUserCredits   (#478)
```

Two problems:

1. **Ordering** — this migration would have been inserted into the middle of a history that a database may already have applied.
2. **The snapshot** — `Migrations/Configuration.cs` sets `AutomaticMigrationsEnabled = true`, so on startup EF compares the current entity model against the **newest** `__MigrationHistory` row's `Model`. With the ids above, that newest row is `AgentCostUserCredits`, whose snapshot knows nothing about the DLP entities. Verified directly by decompressing its `Target`: `dlp_policies = False`. EF would therefore have generated an automatic migration trying to `CREATE` the DLP tables, failing with `There is already an object named 'copilot_dlp_events'` — which was reproduced locally before the fix.

### What was done

Renumbered to **`202609100900001_DlpCopilotImpact`** so it sorts last, and regenerated its model snapshot to describe the **combined** model. Updated in lockstep:

- the four migration files (`.cs`, `.Designer.cs`, `.resx`, `.manual.sql`);
- `IMigrationMetadata.Id`;
- the `Compile` / `EmbeddedResource` / `DependentUpon` entries in `Entities.csproj`;
- the manual script's `@newId`, its `@prevId` and its prerequisite check (now `AgentCostUserCredits`);
- `UrlFullUrlMigrationPipelineTests.LatestId`.

The snapshot was **not** hand-edited. It was produced by serialising the live model with `EdmxWriter` and compressing it exactly as EF does (gzip + base64), after first proving that format by round-tripping an existing migration's `Target`. One detail worth knowing if this ever needs repeating: a `StringWriter` reports UTF-16, so `XmlWriter` emits `encoding="utf-16"` while the blob is stored as UTF-8 **bytes** — EF then fails with `There is no Unicode byte order mark. Cannot switch to Unicode.` The writer must report UTF-8 to match EF's own output.

Both the `.resx` `Target` and the manual script's embedded `@b64` carry the identical regenerated blob, and the stamp writes that blob rather than copying the predecessor's `Model` (correct, because this migration adds entities). The pre-stamp guard remains **schema-only** — it checks the five DLP tables exist and does not inspect data state.

### Manual-SQL run order is now load-bearing

For anyone upgrading a database by hand, the chain is strict and each script hard-fails (`RAISERROR` severity 16) if its predecessor is not stamped:

```
202608310800001_ColumnstoreUsageReportMetrics
  -> 202609090519337_AgentCostTables
  -> 202609090647151_AgentCostUserCredits
  -> 202609100900001_DlpCopilotImpact
```

The first script an existing deployment needs depends on where it already is; the predecessor of the first new one is `ColumnstoreUsageReportMetrics`. This belongs in the eventual `dev`→`main` release notes, along with uploading each `.manual.sql` to the release (CI uploads `**/*.zip` only, so these are always missed).

### Verification

The local test database was dropped and the whole migration chain replayed from scratch: `UrlFullUrlMigrationPipelineTests` 3/3 pass, including the assertion that the database ends at the latest migration id. Full CI (`test_dotnet (Release)`) then passed on the merged tree, and `dev`'s tree is byte-identical to the tree CI validated.

Issue #476, opened 2026-09-09
User-level agent cost attribution: available for Copilot Studio, impossible for Azure


sambetts:
## Update — this is now PARTLY POSSIBLE, and is implemented for Copilot Studio

**Microsoft added per-user entitlement routes to the Power Platform licensing API in July 2026**, after the original research for this milestone was done. Verified against the REST reference:

- [`GET /licensing/entitlements/{entitlementId}/users`](https://learn.microsoft.com/en-us/rest/api/power-platform/licensing/entitlement-insight/get-tenant-users) — returns `TenantCapacityConsumptionUserSnapshotResponseModel` with **`userId`**, `consumed`, `asOfDate`, `environmentId`, `tenantId`, `unit`
- [`GET /licensing/entitlements/{id}/resources/{resourceId}/users`](https://learn.microsoft.com/en-us/rest/api/power-platform/licensing/entitlement-insight/get-tenant-user-consumption-by-resource)
- [`GET /licensing/entitlements/{id}/users/{userId}/resources`](https://learn.microsoft.com/en-us/rest/api/power-platform/licensing/entitlement-insight/get-tenant-resource-consumption-by-user)

So **Copilot Studio credits can now be reported per person, from Microsoft's own figures** — nothing apportioned, nothing inferred. This is implemented: `copilot_studio_credit_user_daily`, plus a "Who is spending the credits" panel on the Agent costs page.

Caveats kept deliberately visible in the UI and the code:

- It is a **separate endpoint**, not a breakdown of the per-agent view, so the two totals are not guaranteed to reconcile exactly.
- The routes are new, so a tenant whose API does not expose them gets a clear message and the per-agent figures still import. The importer treats a 404 as "unavailable", not as a failure.
- `userId` is stored **raw, with no foreign key to `dbo.users`**. Microsoft documents it only as `string` — whether it is an Entra object id, a UPN or something else is not stated, so joining it would rest on an assumption. See the follow-up below.

## Still NOT possible: per-user Azure cost

| Source | Per-user? | Notes |
|---|---|---|
| Power Platform licensing — per-user routes | ✅ **Yes** | Implemented. Documented since July 2026. |
| Power Platform licensing — per-agent route | No | Returns a distinct-user **count** only. |
| Azure Cost Management (all surfaces, incl. full cost-details export) | ❌ **No** | Resource-scoped. No user identity field on any row. This is a property of Azure billing, not a gap in the API. |
[...]

Issue #475, opened 2026-09-09
Portal: granular Agent costs report page for M365 admins


sambetts:
## Goal

An "Agent costs" page in the portal for M365 admins, showing the most granular credit reporting the source data supports.

Delivered at `/insights/agent-costs`:

- **Summary tiles** — credits billed, credits not charged, agents with spend, environments, busiest slice, unclassified-harness credits, tenant capacity headroom and status, Azure spend per billing currency.
- **Daily trend** of billed credits.
- **8-way pivot** — agent, environment, harness, billing feature, AI model, tool invoked, knowledge source, channel. Each with credits, share, days active and peak users.
- **"Every billed line"** — the full billing tuple, one row per day x agent x dimension, sortable and paged, with a CSV export of exactly what is on screen.
- **Azure spend** broken down by meter / service / meter category / resource / resource group / subscription, with an estimate-vs-invoiced flag.
- **Filters** — agent, environment, harness, feature, model, agent-name search. Filter options are read from the data in the window, so a picker never offers a dead end.

## Honesty rules baked into the page

These are the parts most likely to be got wrong later, so they are stated explicitly:

- **There is no cost-per-user column and there never can be.** Microsoft bills Copilot Studio per environment and agent; Azure billing is resource-scoped. The page says so in a standing message rather than leaving an admin to wonder why it is missing.
- **"Busiest slice" is a MAX, never a SUM.** The API reports a distinct-user count *per slice*, and the same person appears under every agent they used, so these counts cannot be added. Adding them would report more "users" than the tenant has. There is an integration test asserting MAX over SUM specifically.
- **Harness is an inference.** Credits whose feature name is not recognised surface as "Unclassified harness" rather than being absorbed into a real harness.
- **Azure figures are marked estimate vs invoiced**, because Azure keeps amending an open billing period for a few days after it ends.
- **A filter that matches nothing looks identical to no spend**, so the page says which filter is active when a query returns zero rows.

## Notes

- `agent_cost_import_log` distinguishes "nothing to import" from "the API refused us" — the two failure modes that otherwise look identical. Errors surface on the page rather than only in the trace log.
- Costs in different currencies are never summed.

sambetts:
Shipped to customers in **Stable build 1830** (`main` @ `cb66e75b`). Closing manually: the fixing PR targeted `dev` and used *Addresses #N*, and since the default branch is `main` a `Closes` keyword there would have been inert.

Issue #474, opened 2026-09-09
Import daily Azure spend via Cost Management (and why Cowork is not an Azure meter)


sambetts:
## Goal

Import daily Azure spend so agent workloads billed directly to an Azure subscription can be reported alongside the rest of the data.

## The Cowork finding (important)

The original ask was specifically for **Microsoft Cowork** costs. Research against Microsoft Learn found:

> *"Looking for information on Copilot Cowork or Work IQ API billing? These services are managed through the Microsoft admin center. See Usage-based Billing and Cost Management for Copilot Credits."*
> — [Meters for Microsoft Copilot Pay-as-You-Go Services](https://learn.microsoft.com/en-us/microsoft-365/copilot/pay-as-you-go/meters)

So Cowork is billed via **Copilot Credits managed in the M365 admin centre**, not an Azure meter. Microsoft publishes **no** Azure resource provider / meter name for Cowork PAYG consumption. The per-user Copilot Credits report in the admin centre exists, but **only as a UI / manual CSV export — there is no public API**.

Consequence: shipping a "Cowork cost importer" with a hard-coded meter would be shipping a guess. Instead this is a **general Azure Cost Management importer** with a configurable scope and meter filter, which is useful regardless and survives whatever Microsoft eventually names the meter.

## Design

- Cost Management **`query`** API (`POST {scope}/providers/Microsoft.CostManagement/query?api-version=2026-06-01`), daily granularity, grouped by subscription / resource / resource group / service / meter category / meter sub-category / meter. Not `generateCostDetailsReport`: that is an async blob-export job for full line-item extracts, and this import wants one aggregated row per day per meter.
- **Configurable filter.** `AzureCostScopes`, `AzureCostMeterFilterDimension`, `AzureCostMeterFilterValues` App Service application settings. With no filter set, every meter at the scope is imported — which is exactly what an operator needs in order to discover what their own agent spend is called.
- **No per-user attribution.** Azure billing is resource-scoped; no Cost Management surface carries a user identity, including the full cost-details export.
- **Costs restate.** Azure re-estimates an open billing period several times a day and finalises it only after the invoice, so the importer re-reads a trailing window (default 5 days) and upserts. `is_estimated` records whether the figure was still provisional when read.
- **Multi-currency safe.** `cost` is always stored with its `currency`, and the reporting layer groups by currency rather than summing across them.

## Delivered

- `azure_cost_daily` table
[...]

Issue #473, opened 2026-09-09
Import billed Copilot Studio credit consumption (MCSMessages)


sambetts:
## Goal

Import **billed** Microsoft Copilot Studio consumption (Copilot Credits) so admins can see what Microsoft actually charged, per agent and per day — as opposed to the per-conversation credit *estimate* this product already derives from Copilot audit events.

## Source

Power Platform licensing API, `MCSMessages` entitlement:

- `GET https://api.powerplatform.com/licensing/entitlements/MCSMessages?api-version=2024-10-01` — tenant entitlement / capacity.
- `GET https://api.powerplatform.com/licensing/entitlements/MCSMessages/resources?fromDate=&toDate=&includeFields=users,tags,asOfDate&pageSize=&continuationtoken=&api-version=2024-10-01` — per-agent consumption.

Grain: usage day x environment x agent x (feature, channel, LLM model, tool, knowledge source).

## Findings that shaped the design

- **No per-user data exists.** Microsoft documents Copilot Studio as *"billed at the environment and agent level, not at the user level"*, and the API returns `metadata.Users` as a **scalar count**, never identities. The importer stores that count and nothing more.
- **Harness is not a reported field.** The admin centre shows "Standard/Copilot Chat harnesses" and "GitHub Copilot harness", but every harness bills through the same `MCSMessages` entitlement. Harness is inferred from the feature name (`Process Agent` => GitHub Copilot), which is community-observed rather than documented — so an unrecognised feature lands on `Unknown` rather than being guessed into a harness.
- **Application-only auth is unconfirmed.** The Power Platform API supports service principals via tenant-scoped RBAC, but Microsoft has not confirmed that the *licensing entitlement routes* accept a client-credentials token. A 401/403 is therefore turned into a specific instruction naming the role assignment, and a failed run does not stamp the cadence gate.
- **Two response envelopes.** The REST reference documents a flat `value[]`; a live tenant with `includeFields` returns the rows nested under `value[].resources[]`. Both are parsed.
- **History restates**, so the importer re-reads a trailing window and upserts.

## Delivered

- `copilot_studio_credit_daily`, `copilot_studio_credit_capacity`, `agent_cost_import_log`
- `CopilotStudioCredits` installer toggle (opt-in, off by default)
- Ports + adapters: `ICopilotStudioCreditSource` / `PowerPlatformLicensingCreditSource`, `IAgentCostStore` / `SqlAgentCostStore`
[...]

PR #478, opened 2026-09-09
Agent cost reporting: Copilot Studio credits (per agent and per user) and Azure spend

sambetts:
Adds agent cost reporting: what Microsoft charges for Copilot Studio agents, and what agent workloads cost in Azure. Two optional imports (off by default), a new SQL schema, and an **Agent costs** page in the portal.

Addresses #473, #474, #475, #476 (milestone: *Agent cost reporting: Copilot Studio credits & Azure-billed agent spend*).

---

## Scope

| Part | Source | Grain |
|---|---|---|
| Copilot Studio credits — per agent | `api.powerplatform.com/licensing/entitlements/MCSMessages/resources` | day x environment x agent x (feature, channel, model, tool, knowledge source) |
| Copilot Studio credits — **per user** | `.../MCSMessages/users` | day x user x environment |
| Copilot Studio capacity | `.../entitlements/MCSMessages` | daily tenant snapshot |
| Azure spend | Cost Management `query` API | day x (2 configurable dimensions) |

New tables: `copilot_studio_credit_daily`, `copilot_studio_credit_user_daily`, `copilot_studio_credit_capacity`, `azure_cost_daily`, `agent_cost_import_log`.

New toggles: `CopilotStudioCredits`, `AzureCostManagement`. `CONFIG_VERSION` 2.2.0 → **2.3.0**.

---

## Two research findings that shaped the design

**1. Cowork is not an Azure meter.** The original ask was for Cowork costs specifically. Microsoft's own meters documentation routes it away from Azure:

> *"Looking for information on Copilot Cowork or Work IQ API billing? These services are managed through the Microsoft admin center."*
> — [Meters for Microsoft Copilot Pay-as-You-Go Services](https://learn.microsoft.com/en-us/microsoft-365/copilot/pay-as-you-go/meters)

Cowork bills through Copilot Credits in the M365 admin centre, and Microsoft publishes **no** Azure resource provider or meter name for it. The admin-centre per-user report is **UI/CSV only — no public API**. So this ships as a **general Azure Cost Management importer** with a configurable scope and meter filter rather than a "Cowork importer" with a guessed meter constant. With no filter set it imports every meter at the scope, which is how an operator discovers what their own agent spend is actually called. See #474.

**2. Per-user Copilot Studio credits became possible in July 2026.** Microsoft added per-user entitlement routes ([`GET /licensing/entitlements/{id}/users`](https://learn.microsoft.com/en-us/rest/api/power-platform/licensing/entitlement-insight/get-tenant-users), returning `userId` + `consumed` + `asOfDate`). These are imported. **Nothing is apportioned or inferred** — the figures are Microsoft's. Azure spend still has no per-user view and cannot have one; that caveat is now stated separately per source rather than applied to both. See #476.

---

## Architecture

Ports and adapters throughout, so the logic is testable without HTTP or SQL:

- `ICopilotStudioCreditSource` / `PowerPlatformLicensingCreditSource`, `IAzureCostSource` / `AzureCostManagementSource`
- `IAgentCostStore` / `SqlAgentCostStore`, `IAgentCostReportStore` / `SqlAgentCostReportStore`
- Pure parsers: `CopilotStudioCreditParser`, `AzureCostQueryParser`, `CopilotStudioHarnessClassifier`, `AgentCostRowHasher`

`AgentCostImportPhase` is deliberately **not** an `IGraphImportSection`: these imports authenticate to `api.powerplatform.com` and `management.azure.com`, different audiences with different role requirements from Graph. Importer construction is lazy, so a deployment with the toggles off never builds a client or acquires a token.

**Both sources restate history** (Azure re-estimates an open billing period several times a day; Copilot Studio recalculates as usage days settle), so both importers re-read a trailing window and **upsert** on `(usage_date, hash)`. The unique indexes are the upsert keys, not an optimisation.

---

## Migrations — both ADDITIVE

`202609090519337_AgentCostTables` and `202609090647151_AgentCostUserCredits`. Five new empty tables and three unique indexes. Nothing existing is altered, dropped or rewritten and no existing query changes plan, so **there is no "before" state to measure** and both are out of scope for the prove-it performance rule. Upgrade time is independent of tenant size; no maintenance window needed.

Both `.manual.sql` scripts were **validated end to end against LocalDB as a chain**: EF-migrated to the predecessor, applied in migration-id order, then asserted 5/5 tables, 3/3 unique indexes, every text column `nvarchar`, both `__MigrationHistory` stamps byte-identical to their own migration snapshots, both scripts idempotent on re-run, and EF reporting **0 pending migrations**. Stamp guards check **schema only** — no data-state guard.

---

## Review

Three models reviewed this (`claude-opus-4.8`, `gpt-5.6-sol`, `gemini-3.8-flash`), and two API-contract claims were then verified against Microsoft's REST reference. That caught **two defects that would have stopped the feature working entirely**:

1. **The Azure query sent 7 `grouping` clauses.** `QueryDataset` declares `grouping` with `maxItems: 2` in *every* API version — every request would have returned HTTP 400, on every run, for every customer. Now configurable, capped at 2, defaulted to `ResourceId` + `Meter` (subscription and resource group are parsed back out of the resource id). Regression test asserts the cap.
2. **Copilot consumption was read as one 7-day request** and split by the undocumented `asOfDate`. The endpoint aggregates over the range given, so a week's spend would have landed on one day. Now one request per day. Regression test asserts it.

Also fixed: usage-date clamping that duplicated spend as the window slid; installer checkboxes overlapping a label; a 10-minute retry storm on an unauthorised/unconfigured import; partial paging stored as success; an Azure refresh window narrower than its own "still provisional" rule; invalid dates silently answering for the default period; pivot shares computed over the visible top-20; CSV formula injection; quantities and sub-unit money rounded to zero; `"Invoiced"` claiming more than the flag means; raw enum leaked into the UI.

---

## Honesty rules baked in

These are the things most likely to be quietly broken later, so they are asserted in tests:

- **Distinct-user counts are a MAX, never a SUM.** The per-agent API reports a count *per slice*; the same person appears under every agent they used, so adding them would report more users than the tenant has.
- **Costs are never summed across currencies**, and Azure top-N is taken per currency — ranking across them compares numbers that are not comparable.
- **Harness is an inference** from a community-observed feature name; anything unrecognised surfaces as "Unclassified harness" rather than being absorbed into a real harness.
- **An import that is refused looks nothing like an import that found nothing** — `agent_cost_import_log` separates them, and the page shows the reason.
- A filter that matches nothing is called out, because it is indistinguishable from no spend.

---

## Tests

- **57** agent-cost unit + SQL integration tests (parsers incl. both envelopes, hashing, aggregation, upsert idempotency under a re-read, Unicode round-trip to SQL, paging guards, provisional-cost rule, per-user mapping, route-unavailable handling)
- **68** migration-pipeline / toggle-coverage / health tests (includes the `LatestId` assertion)
- **15** portal tests

`test_dotnet (Release)` gates this PR. The four VNet install tests fail locally without a real `InstallerTestConfig.json`; unrelated and pre-existing.

---

## Reviewer hotspots

- `AzureCostManagementSource.BuildRequestBody` — the 2-grouping limit is a hard API constraint, not a preference.
- `CopilotStudioCreditImporter.ResolveUsageDate` — the date must stay a fact about the *request*; anything else reintroduces the duplication bug.
- `AgentCostImportPhase.StampOrRetry` — deliberately asymmetric: transient failures retry next cycle, authorisation failures back off.
- `SqlAgentCostReportStore` — all aggregation is server-side; check nothing has become client-side evaluation.

## Deferred

- Verifying the format of `userId` so per-user rows can join `dbo.users` and show real names (#476).
- Cowork per-user credits, pending an API (#474).
- No wiki changes in this PR.

---

## Review loop (this is why the diff has five commits)

Four rounds across five models, each round re-reviewing the previous round's fixes. Rounds 2 and 3 each found defects **introduced by the previous round's fix** — the exact pattern `release-critic.agent.md` exists for. Round 4 came back clean.

| Round | Found |
|---|---|
| 1 | Azure query sent 7 `grouping` clauses (API caps at 2 → HTTP 400 on **every** run); Copilot consumption fetched as one 7-day request and split by an undocumented field; usage-date clamping that duplicated spend as the window slid; installer checkboxes overlapping a label; 10-minute retry storm; partial paging stored as success; CSV formula injection; several over-claiming labels |
| 2 | **Both of round 1's own fixes were wrong.** A quiet day (204) aborted the whole per-user window *and* discarded the days already read, then reported success. And a transient failure occurring alongside a refusal was misclassified as "refused" and suppressed for a full day — at both cadence call sites |
| 3 | Azure costs were written as a **delta** when the query returns a **snapshot** (narrowing the filter kept old rows; changing the grouping doubled the total); a closing billing period never got its one final read, so its rows stayed "estimate" for ever; availability ignored the per-user and capacity logs; the chart computed its own share from the visible top-20, contradicting the table beside it |
| 3b (self-found) | Making the write destructive raised the cost of any silent partial read from "stale data" to **deleted data** — one path (empty body mid-paging) still returned what it had. Now throws |
| 4 | **Clean.** Two residual observations, both assessed as non-regressions and left deliberately (below) |

Two API-contract claims were verified against Microsoft's own REST reference rather than taken on trust — the `maxItems: 2` grouping cap and the range-vs-per-day semantics of the licensing endpoint. Both turned out to be real, and both would have shipped a feature that never worked.

**Tests: 66** agent-cost (unit + SQL integration), **68** migration/toggle/health, **15** portal. Three of the SQL tests exist solely to bound the new delete: it must remove superseded rows, clear the window when the query matches nothing, and never touch another scope or another day.

## Deliberately not fixed

- **One shared error slot for three portal panels.** If two panels fail simultaneously and one recovers via its own trigger, the other's banner can be cleared early. Round 3 strictly improved this (a sibling's success no longer wipes another panel's error); a full fix needs per-panel error state and was judged not worth the churn.
- **Per-user/capacity messages defer while the per-agent import is also failing** — intentional "report once, not twice"; the per-agent banner still shows, so nothing looks healthy when it isn't.

## Live-tenant verification (added after review)

The licensing API was exercised against a real tenant with an administrator token. No tenant values
are recorded here or in the tests — only response *shapes*, with zeroed GUIDs.

| Route | Administrator token |
|---|---|
| `/licensing/entitlements/MCSMessages` | 200 |
| `/licensing/entitlements/MCSMessages/users` | 200 |
| `/licensing/entitlements/MCSMessages/resources` | **403** |

Confirmed and now pinned by tests built from the captured shapes:

- **`userId` is an Entra object ID (GUID), not a UPN.** It does rule out a join to
  `dbo.users.user_name` — but **not** a join to `dbo.users`. See the correction below.
- The per-user envelope is the nested `value[].users[]` form and **`asOfDate` is present without
  `includeFields`** — both already handled.
- Per-user rows carry `metadata.NonBillableQuantity`; a user can have `consumed = 0` with
  `NonBillableQuantity = 1`. Parsed so it is not lost at the boundary, but **not persisted** — the
  per-user table has no column for it. Follow-up on #476.
- Tenant entitlement returns `payGo.entitled` with **no** `payGo.consumed`, so `PayAsYouGoConsumed`
  stays null. It deliberately does not fall back to `entitled` — different quantities.

**Reviewer hotspot:** `/resources` refused a *Global Administrator's* token while two sibling routes
accepted it, so that route needs something beyond directory admin rights. Unresolved; tracked on #476
and called out in the wiki so admins do not read it as a misconfiguration on their side.

This probe also caught a real defect, fixed in this PR: a persistent HTTP 500 on the per-user route was
classified transient, so the cadence gate never stamped and the **whole** import would have re-run every
cycle forever. The per-user read is now best-effort and excluded from the cadence decision.
---

## Correction: per-user credits now carry a REAL foreign key to `dbo.users`

An earlier revision of this PR (and issue #476) claimed that linking per-user credits to a person needed a
new object-ID column on `dbo.users` plus a Graph backfill. **That was wrong.** `dbo.users.azure_ad_id`
already holds the Entra object ID — the user import maps it straight from Graph's `user.id`
(`UserMetadataMappingRules.BuildPlan`), which is exactly the value the licensing API returns as `userId`.
No new column and no backfill were required.

### Schema

`copilot_studio_credit_user_daily.user_id` is now a real FK to `dbo.users.id`; the raw identifier moved to
`entra_object_id`. **Both migrations are reshaped in place rather than superseded**, because this PR is
unmerged and those tables have never run on a customer database — so there is no upgrade path to preserve.

The key is **nullable and does not cascade**:

- **Nullable** because a billing row must survive a user who cannot be resolved (deleted from the
  directory, or created since the last user import). Such a row still reports its spend against the raw
  object ID and is re-attributed later. Losing money to keep a key tidy is the wrong trade on a cost report.
- **`NO ACTION`** because deleting a user must never silently erase spend history and change historical
  totals. Verified that production code never deletes users (only tests do), so this can only ever surface
  as a loud failure rather than quiet data loss.

Still **purely additive**: the FK points *at* `dbo.users` but adds no column to it and validates no existing
row, because the referencing table is new and empty.

### Resolution

Two steps, cheapest first — everything already in `dbo.users` is matched in one query against
`azure_ad_id`, and only the remainder costs a directory call.

- **Budgeted** at 200 directory lookups per run, so a tenant that has never run the user import cannot turn
  the first agent-cost import into one Graph call per user.
- **"No such user" and "could not ask" are tracked separately** — the first is settled and never retried,
  the second is retried next cycle. Conflating them either abandons a resolvable user or re-asks about a
  deleted one for ever.
- **`EnsureUserAsync` matches on UPN first**, then inserts. Users arrive in `dbo.users` from several places
  — the audit-activity staging merge inserts them by user name with *no* object ID — so matching on the
  object ID alone would have created a duplicate row for someone already present. When it finds such a
  user it backfills their `azure_ad_id`, permanently removing the need for a directory call for them.
- **A re-attribution pass** covers rows outside the trailing window, which the import never re-reads.
  Without it, a row whose user was created five minutes too late would stay unattributed for ever.
- Uses the Graph client the user import already builds, so **no new permission or consent**.
- Nothing here can fail the import: every path degrades to an unlinked row that still reports its spend.

Ports and adapters as elsewhere in this feature: `IEntraUserLookup` / `GraphEntraUserLookup`,
`IAgentCostUserLinkStore` / `SqlAgentCostUserLinkStore`, with `AgentCostUserResolver` as pure orchestration
testable without HTTP or SQL.

### Two defects found while building this

1. **The manual SQL upgrade scripts could not stamp `__MigrationHistory` under `sqlcmd`.** The stamp
   rebuilds the model blob with an XML method, which requires `QUOTED_IDENTIFIER ON`. SSMS connects with it
   on; **`sqlcmd` defaults it off**. The schema work succeeded and then the stamp failed — leaving the
   migration unstamped and, because the scripts form a prerequisite chain, **blocking the next script**.
   Fixed in all three scripts that use that mechanism (including the already-shipped
   `202608190725064_AddCopilotUsageReports`). The previous validation harness never invoked `sqlcmd`, which
   is why this was not caught.
2. **The top-users report split one person into two rows** while some of their rows were linked and some
   were not — "Alice, 2 credits" beside "Unresolved, 3 credits". It now groups on the Entra object ID (the
   stable identity) rather than on the link, which changes as re-attribution runs.

### Also in this push

**Every agent-cost API model was serialising in PascalCase.** The Web project registers no camelCase
contract resolver, and `AgentCostReportModels.cs` had no `[JsonProperty]` attributes on any of its 85
response properties, so the portal read every field as `undefined` and the page died at the first `.map`
with `Uncaught TypeError: Cannot read properties of undefined (reading 'map')`. TypeScript could not catch
it — the response is typed by assertion at the fetch boundary, not validated. Added the attributes plus
`AgentCostApiContractTests`, which serialises each model and asserts the exact JSON key set against names
copied from `types/agentCosts.ts` (deliberately not derived by camelCasing the C# name — a rule computed
from the code under test would agree with a rename on either side and prove nothing).

### Verification

- EF model snapshot regenerated. A full `Update()` with the repository's own configuration applies cleanly
  and then scaffolds an **empty** migration — model and migrations agree.
- Manual scripts validated as a chain **through `sqlcmd`**: applied from the predecessor state, idempotent
  on re-run, correct columns / indexes / FK, both `__MigrationHistory` stamps byte-identical to their own
  `.resx` `Target`, and EF reporting **zero pending migrations**.
- The `.ts` interfaces were diffed programmatically against the `[JsonProperty]` names: **12/12 interfaces,
  85/85 fields match**.
- **99 agent-cost tests, 132 portal tests, `tsc` clean.** Both new report tests were confirmed to fail
  without their fix, and the contract guard was confirmed to fail when an attribute is removed.

### Reviewer hotspots

- `SqlAgentCostUserLinkStore.GetUserIdsByObjectIdAsync` — `azure_ad_id` is `nvarchar(max)` and so cannot be
  indexed, making each chunk a scan of `dbo.users`. The filter is pushed into SQL rather than loading the
  table and matching in memory: one scan returning a handful of rows beats transferring 200,000. Acceptable
  for a once-a-day import; making it a seek would mean narrowing the column, which is a measured change in
  its own right and deliberately not bundled here.
- `AgentCostUserResolver.ResolveAsync` — the budget and the not-found/failure split are both load-bearing.
- `SqlAgentCostReportStore.GetTopUsersAsync` — grouping on `entra_object_id` rather than on `user_id` is
  what stops one person appearing twice.


sambetts:
**Post-merge note: the migration ordering around this PR changed during integration into `dev`.**

This PR's migrations were left exactly as authored, but #477 (DLP) added a migration whose id fell **between** them:

```
202609090519337_AgentCostTables        (this PR)
202609090601123_DlpCopilotImpact       (#477)  <- interleaved
202609090647151_AgentCostUserCredits   (this PR)
```

That was resolved on #477's side by renumbering its migration to `202609100900001_DlpCopilotImpact` so it sorts **after** both of this PR's migrations, and regenerating its model snapshot to describe the combined model. Nothing in this PR's migrations, snapshots or manual scripts was modified.

One consequence worth recording here: `202609090647151_AgentCostUserCredits` is no longer the newest migration, so it is no longer the row EF compares the entity model against on startup. Its own snapshot is unchanged and still correct for its point in the chain.

The manual-SQL upgrade chain is now, in order:

```
202608310800001_ColumnstoreUsageReportMetrics
  -> 202609090519337_AgentCostTables
  -> 202609090647151_AgentCostUserCredits
  -> 202609100900001_DlpCopilotImpact
```

Each script hard-fails (`RAISERROR` severity 16) if its predecessor is not stamped in `__MigrationHistory`, so operators upgrading by hand must run them in that order. All four `.manual.sql` files need uploading to the eventual stable release — CI uploads `**/*.zip` only.

Full detail of the renumber is on #477. Verification: the local test database was dropped and the whole chain replayed from scratch (`UrlFullUrlMigrationPipelineTests` 3/3), and CI `test_dotnet (Release)` passed on the merged tree.

PR #480, opened 2026-09-09
Resolve audit user ids that are Entra object ids to UPNs before they become users

sambetts:
## Summary

The importer stages the Management Activity API's `UserId` straight into `dbo.users.user_name`, and the merge creates a user row for any value it has not seen:

```sql
INSERT INTO users(user_name)
SELECT DISTINCT imports.user_name ... WHERE users.user_name IS NULL;
```

That is correct for a UPN, which is nearly always what arrives. But **the common schema does not guarantee one**, and there was no guard anywhere in the path — `ActivityStagingPass` passes `log.UserId` through verbatim. Microsoft documents the same field carrying **Entra object ids (GUIDs)**, **Windows SIDs** and **service principals** such as `app@sharepoint`; DLP records additionally carry the literal `DlpAgent`.

So each distinct non-UPN value silently becomes its own "user". One person seen under an object id splits across two rows, and every per-user report under-counts them — with nothing in the logs to say so.

## What this does

Resolve an Entra object id to the person's UPN **before** it becomes a user row, so it lands on their existing `dbo.users` record.

| Piece | Role |
|---|---|
| `AuditUserIdentity` | Pure classifier: UPN / Entra object id / system account / unrecognised. |
| `AuditUserIdentityResolver` | Database-first resolution, Entra on a miss, caching and write-back. |
| `SqlUserEntraIdStore` | Reads and fills `users.azure_ad_id`. |
| `GraphEntraUserLookup` / `LazyGraphEntraUserLookup` | Directory lookup, with the client built only if needed. |

**Only Entra object ids are resolved.** A UPN needs nothing; a service principal is not a person; and resolving an unrecognised value would be guesswork. `app@sharepoint` is specifically handled because it *contains an `@`* — a naive UPN test would let it straight through to the directory.

**Database first, and usually only.** The Graph user import already stores every user's `azure_ad_id`, so on a tenant that runs it the database answers everything and Entra is never called. What Entra does teach us is **written back** to `azure_ad_id`, so the next cycle answers from SQL.

## Performance — free on the hot path

This sits in the audit save path for **every** workload, so it is built to cost nothing in the normal case:

- A batch whose `UserId`s are all UPNs costs **one character test per event** and never touches the resolver, the database or Graph. There is a test asserting exactly that (`SavePipeline_AllUpns_NeverCallsTheResolver`).
- Resolution is **batched per staging pass**, not per event.
- Outcomes are cached for the run **including failures**. Without the negative cache, a deleted user — which by definition can never resolve — would be looked up again for every event that mentions it: the per-event Graph call this repo's guidance forbids at the ~200k-user target.
- `MaxLookupsPerBatch` (200) is a backstop so a pathological batch degrades into unresolved ids rather than a long stall.

## Every failure mode degrades to today's behaviour

Nothing here can drop an event or invent an identity. An unresolvable id, a directory outage, a failed credential, and a failed write-back **all keep the raw value**, exactly as before this existed. Each is covered by a test.

## Verification

- Full solution builds.
- **29 focused tests pass** — 13 on classification and resolution policy, plus 4 new end-to-end assertions through `ActivityStagingPass` proving a staged row actually carries the resolved UPN (and that an unresolvable id keeps its raw value).
- **Broad regression run: 989 tests, 980 passed, 5 failed.** All 5 failures are pre-existing environmental ones — `AADSTS90002: Tenant '00000000-…-000000000002' not found` and Redis RBAC — in Teams/calls tests that need real Azure credentials. None touch this code.

No schema change, no migration, no new permission: `azure_ad_id` already exists and `User.Read.All` is already required.

## Reviewer hotspots

1. **`AuditUserIdentity.Classify` ordering** — the system-account check runs *before* the `@` test, which is what stops `app@sharepoint` being treated as a UPN and sent to Entra.
2. **The group filter now tests the resolved UPN**, not the raw id. Before, an object id would have been tested against `UserGroupsFilter` and failed, silently dropping that person's events on a scoped tenant. This is a behaviour change, and I believe it is the correct one.
3. **Write-back only fills a gap** — an existing `azure_ad_id` is never overwritten on the strength of an audit record; the Graph user import stays the authority for that column.

## Honest caveat on the evidence

I could not reproduce this from data. I checked both local databases and found **zero** GUID-shaped `user_name` rows (the non-UPN ones there are all stress-test artifacts like `test chat user 6392…`). What is proven is the **code path** — there is no guard, so any non-UPN value becomes a user row — and Microsoft's documented schema for the field. Treat this as a correctness guard rather than a fix for an observed incident; that is also why it is built to be free when the case never fires.

## Deferred

Repairing junk rows already present in `dbo.users`. That is a data migration with its own risks (merging user rows, re-pointing foreign keys) and belongs in its own change, once there is evidence of rows that need it.

### DIFF
--- a/src/AnalyticsEngine/Tests.UnitTests/ActivitySavePipelineTests.cs
+++ b/src/AnalyticsEngine/Tests.UnitTests/ActivitySavePipelineTests.cs
@@ -78,6 +78,33 @@ namespace Tests.UnitTests
         private static ActivityStagingPass PassFor(AuditFilterConfig filterConfig, ILogger logger)
             => new ActivityStagingPass(filterConfig, GroupsCache(), new UserGroupsFilterModel("Finance"), logger);
 
+        private static ActivityStagingPass PassFor(AuditFilterConfig filterConfig, ILogger logger, IAuditUserIdentityResolver resolver)
+            => new ActivityStagingPass(filterConfig, GroupsCache(), new UserGroupsFilterModel("Finance"), logger, resolver);
+
+        /// <summary>Resolver stub returning a fixed object-id to UPN map, and counting its calls.</summary>
+        private sealed class StubUserIdentityResolver : IAuditUserIdentityResolver
+        {
+            private readonly Dictionary<string, string> _map;
+            public int Calls { get; private set; }
+
+            public StubUserIdentityResolver(Dictionary<string, string> map = null)
+            {
+                _map = map ?? new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            }
+
+            public Task<IReadOnlyDictionary<string, string>> ResolveToUpnAsync(IReadOnlyCollection<string> ids)
+            {
+                Calls++;
+                var hits = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                foreach (var id in ids)
+                {
+                    string upn;
+                    if (_map.TryGetValue(id, out upn)) hits[id] = upn;
+                }
+                return Task.FromResult<IReadOnlyDictionary<string, string>>(hits);
+            }
+        }
+
         private sealed class SequenceActivityImportCacheProvider : IActivityImportCacheProvider
         {
             private readonly Queue<ActivityImportCache> _caches;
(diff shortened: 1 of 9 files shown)


sambetts:
Closing this: it was opened from a request that was meant for a different session (agent costs) and landed in the wrong window. Nothing here was asked for in the context it was built in, so it should not sit in the review queue.

**The branch `sambetts/resolve-audit-user-identity` is left in place**, so nothing is lost and this can be reopened in seconds if it turns out to be wanted on its own merits.

For anyone who finds this later, what it contains and the one caveat that matters:

- It resolves Management Activity API `UserId` values that arrive as Entra object ids to the user's UPN before the merge turns them into a `dbo.users` row. The code path it guards is real - `ActivityStagingPass` passes `UserId` through verbatim and the merge creates a user row for any unseen value, with no guard anywhere.
- **But I could not reproduce it from data.** Both local databases contain zero GUID-shaped `user_name` rows. It is a correctness guard against a documented schema possibility, not a fix for an observed incident - which is also why it is built to cost nothing when the case never fires.

If the agent-costs work needs user-id resolution, it is likely a *different* problem to this one and worth designing against that feature's own records rather than reusing this.

Issue #495, opened 2026-09-10
Daily usage reports: instrument and benchmark save-phase lookup and persistence costs


sambetts:
## Summary

The daily usage-report save boundary combines database reads, per-record lookup resolution, projection/dirty checking and batched EF writes. Existing start/completion messages cannot identify which of these costs dominates at large-tenant scale.

Add privacy-safe stage measurements and a reproducible synthetic-scale benchmark, then address the demonstrated bottleneck. In particular, quantify the cold-cache lookup path rather than assuming the existing batched saves eliminate all per-user database work.

This issue deliberately includes no production duration, row count, query plan, database/resource name or customer identifier. The scale and latency examples below are explicitly synthetic; they are not benchmark results.

## Source-backed boundaries

- [AbstractDailyActivityLoader.SaveLoadedReportsToSql](https://github.com/pnp/Microsoft365-Analytics-Insights/blob/bfa90a818595c1d9cc276fe1057bf4b487886f82/src/AnalyticsEngine/WebJob.Office365ActivityImporter.Engine/Graph/UsageReports/AbstractDailyActivityLoader.cs#L229-L347) loads existing rows for a date, checks scope, resolves lookups, maps report fields, dirty-checks and saves batches.
- `SaveBatchSize` defaults to 1000. Dirty checking already avoids rewriting unchanged rows, and saved-row release/bulk-write handling limits EF tracking overhead. Preserve these improvements.
- [ResolveLookupIdAsync](https://github.com/pnp/Microsoft365-Analytics-Insights/blob/bfa90a818595c1d9cc276fe1057bf4b487886f82/src/AnalyticsEngine/WebJob.Office365ActivityImporter.Engine/Graph/UsageReports/AbstractDailyActivityLoader.cs#L360-L387) first checks `ConcurrentLookupDbIdsCache`; a miss invokes `reportPage.GetOrCreateLookup(lookupCache)`.
- [Usage-report lookup objects](https://github.com/pnp/Microsoft365-Analytics-Insights/blob/bfa90a818595c1d9cc276fe1057bf4b487886f82/src/AnalyticsEngine/WebJob.Office365ActivityImporter.Engine/Entities/Serialisation/UsageReports/Abstract.cs#L19-L40) connect that miss path to the underlying database lookup.
[...]

PR #498, opened 2026-09-10
Instrument usage report save diagnostics

sambetts:

### DIFF
--- a/src/AnalyticsEngine/Benchmarks/UsageReportSaveSyntheticBenchmark.sql
+++ b/src/AnalyticsEngine/Benchmarks/UsageReportSaveSyntheticBenchmark.sql
@@ -1,0 +1,130 @@
+/*
+Synthetic daily usage-report save benchmark harness for issue #495.
+
+Run only against an empty throwaway SQL Server database. This script creates synthetic tables with the
+same relevant production column types and current query shapes used by AbstractDailyActivityLoader:
+  - dbo.users.user_name is varchar(250), matching production Entra UPN storage.
+  - dbo.outlook_user_activity_log uses int identity id, datetime date/last_activity_date, int user_id,
+    bigint report counters, FK to users and the current IX_date shape from IndexUsageReportSnapshots.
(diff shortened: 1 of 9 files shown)


sambetts:
### Validation record correction (post-merge)

The **Validation** section in the PR description above was written during the first build attempt and is stale. It reports the build failing on a missing `Microsoft.WebApplication.targets` and the tests never running. That was an environment problem (VS BuildTools rather than a full VS install), not a problem with this change, and it was resolved before merge.

Actual validation, using the VS 18 Enterprise toolchain:

| Run | Result |
|---|---|
| `MSBuild Tests.UnitTests.csproj` restore + build (Release/anycpu/x86) | passed |
| `DailyActivityLoaderTests` + `UsageReportRefreshPolicyTests` | 26 / 26 passed |
| `FullyQualifiedName~UsageReport|FullyQualifiedName~DailyActivity` | 83 / 83 passed (excluding the live-credential Graph test) |
| Broad regression filter | 185 total, 184 passed, 1 failed |
| `test_dotnet (Release)` on this PR | success |

The single failure is `GraphUsageReportImportTests.SharePointSitesUsageLoaderTest`, which performs a real Entra OAuth call and only passes in CI where repository secrets are substituted into `App.Release.config`. It fails identically on an unmodified `origin/dev` checkout, so it is a pre-existing local-environment failure and not a regression.

### Post-merge verification on `dev`

After this PR and #499 / #500 were merged, `dev` was rebuilt and re-tested locally: **218 total, 217 passed, 1 failed** — again only the live-credential test above. A full `tests.yml` `workflow_dispatch` run on `dev` also completed green (`test_dotnet (Release)`: success), which matters because `tests.yml` deliberately does not run on pushes to `dev`, so three individually-green PR builds would not otherwise have proven the merged result.

The **Deferred intentionally** section remains accurate: the 200k-user SQL measurement run and any resulting lookup-preload / index / migration work are still outstanding for #495.

Issue #520, opened 2026-09-12
BLOCKER: installer config encryption (StringCipher) throws on .NET 10 - Rijndael 256-bit block is unsupported


sambetts:
## Severity

**Blocker for the installer on .NET 10.** Not a warning, not an obsolescence cleanup — on .NET 10 the Control Panel app can neither open an existing configuration file nor save a new one. It fails at the first action an admin takes.

Originally noted only as an obsolete-API warning in #512 / #514. That understated it; this issue supersedes those notes.

## The problem

`Common/DataUtils/StringCipher.cs` encrypts the two secrets stored in installer config files:

- `SolutionInstallConfig.SQLServerAdminPassword` (via `SQLServerAdminPasswordHash`)
- `AppRegistrationCredentials.Secret` (via `SecretHash`)

It uses:

```csharp
using (var symmetricKey = new RijndaelManaged())
{
    symmetricKey.BlockSize = 256;   // <-- the problem
    symmetricKey.Mode = CipherMode.CBC;
    symmetricKey.Padding = PaddingMode.PKCS7;
```

`BlockSize = 256` is **Rijndael, not AES**. AES is Rijndael restricted to a 128-bit block. On .NET Core / .NET 5+, `RijndaelManaged` is implemented over AES and supports only the 128-bit block, so it throws at runtime:

```
System.PlatformNotSupportedException: BlockSize must be 128 in this implementation.
```

This **compiles cleanly** on .NET 10 (only `SYSLIB0022` obsolete warnings), so nothing catches it until the code executes. There is no in-box replacement, because .NET has never shipped a 256-bit-block Rijndael outside the old `RijndaelManaged` on .NET Framework.

## Reproduction — measured, both directions

Harness: one source file (a verbatim copy of the shipping `StringCipher`) multi-targeted `net48;net10.0`. The net48 leg was executed under Windows PowerShell 5.1 (which runs on **.NET Framework 4.8.9310.0**) by loading the compiled assembly; the net10.0 leg under **.NET 10.0.11**.

| Test | Runtime | Result |
|---|---|---|
| Encrypt + round-trip, shipping code | .NET Framework 4.8.9310.0 | ✅ OK (baseline) |
| Round-trip, shipping code | .NET 10.0.11 | ❌ `PlatformNotSupportedException` |
| **Decrypt a net48-encrypted config**, shipping code | .NET 10.0.11 | ❌ `PlatformNotSupportedException` |

[...]

PR #521, opened 2026-09-12
Lock the installer config encryption format with golden-vector tests

sambetts:
## Scope

Adds test coverage for `DataUtils.StringCipher`, which had none. No production code changes — two files: the new test class and one `<Compile>` entry in the legacy `Tests.UnitTests.csproj`.

`StringCipher` encrypts the two secrets persisted in every customer's installer configuration file:

- `BaseSolutionInstallConfig.SQLServerAdminPasswordHash` (the SQL admin password)
- `AppRegistrationCredentials.SecretHash` (the app registration client secret, for both the installer and activity accounts)

The invariant being locked: **a config file must open with the same password that saved it, regardless of which build of the installer opens it.** Nothing enforced that before this PR.

## Why golden vectors rather than round-trips

The six ciphertext constants were captured from the **shipping `DataUtils.dll` executed on .NET Framework 4.8.9310.0**, then hard-coded.

This is deliberate and is the substance of the PR. An encrypt-then-decrypt round-trip test passes on any implementation that is *consistently* wrong — it validates self-consistency, not compatibility. Only a ciphertext produced by a previous build can prove that files already sitting on customer disks still decrypt. A round-trip test would have caught none of the failure modes this PR guards against.

Cases covered: ASCII, non-Latin plaintext, empty string, special characters common in generated secrets, a non-Latin passphrase (exercising UTF8 encoding inside the KDF), and a 500-character multi-block payload (so a CBC chaining or block-size regression cannot hide behind a single-block secret).

> The vectors must not be regenerated. If a change makes them fail, the change has broken compatibility with every config file in the field.

## Format assertions

The cipher is **Rijndael with a 256-bit block**, which is *not* AES — AES is Rijndael fixed to a 128-bit block. The tests assert the wire format explicitly:

- 32-byte salt + 32-byte IV header
- encrypted payload is a whole number of 32-byte blocks
- salt/IV are random per call (a fixed IV would be a real security defect a round-trip test cannot see)
- secrets are never persisted in plain text
- freshly-written ciphertext has the same length as the shipped format for the same plaintext

A switch to AES would produce a 16-byte block and a 16-byte IV, stranding every existing config file. These assertions make that fail loudly and immediately.

## Test list (21)

| Group | Count | Guards |
|---|---|---|
| Golden vectors | 6 | Config files written by the shipping build still decrypt |
| Format invariants | 8 | 256-bit block layout, random IV per call |
| Round-trip | 2 | Encrypt/decrypt symmetry; wrong password rejected |
| Config level | 4 | `LoadFromJson` / `ToJson` across all three secrets; no plaintext persisted |
| Encoding | 1 | UTF8 boundary for non-Latin secrets |

The config-level tests go through `SolutionInstallConfig.LoadFromJson` / `ToJson` rather than calling `StringCipher` directly, so they cover the path the installer actually uses.

## Verification

Built and run with the Enterprise MSBuild + `vstest.console.exe`, matching CI:

```
Passed!  - Failed: 0, Passed: 21, Skipped: 0, Total: 21 - Tests.UnitTests.dll (net48)
```

**Mutation-tested, because a green test proves nothing on its own.** Changing `symmetricKey.BlockSize` from `256` to `128` — simulating exactly the "modernise this to AES" regression these tests exist to stop — produces **14 failures**, including all six golden vectors and every config-level test. Reverting returns 21/21.

## Reviewer notes

- `Tests.UnitTests.csproj` is still a legacy (non-SDK) project, so the new file needs its explicit `<Compile>` entry. Easy to miss.
- The source file is UTF-8 **with BOM** — required for the non-Latin constants to compile correctly under MSBuild.
- All values are synthetic: passphrase `CorrectHorseBattery`, secrets like `P@ssw0rd-SQLAdmin-2026`. No real customer password, secret, tenant or connection string. The golden vectors are ciphertexts of those synthetic values.
- Runtime is ~1s; no database, network or external dependency.

## Related — why this was written now

Found while assessing the .NET 10 port (milestone *.NET 10 parallel branch*). On .NET 10, `RijndaelManaged` is implemented over AES and rejects the 256-bit block with `PlatformNotSupportedException: BlockSize must be 128 in this implementation`, so the installer can neither open nor save a config file. It compiles cleanly, so only a runtime test catches it.

This PR is the guard, landing **before** any port work so there is something to port *against*. It is net48-safe and passes against the current implementation unchanged — it is worth having on `dev` on its own merits regardless of whether .NET 10 ever happens.

A secondary finding for the eventual fix, not addressed here: `SolutionInstallConfig.LoadFromJson` catches only `CryptographicException`, and `PlatformNotSupportedException` does not derive from it — so on .NET 10 the failure would be an unhandled crash rather than the intended "wrong password, `DecryptedOk = false`" path.

Addresses #520.

### DIFF
--- a/src/AnalyticsEngine/Tests.UnitTests/StringCipherCompatibilityTests.cs
+++ b/src/AnalyticsEngine/Tests.UnitTests/StringCipherCompatibilityTests.cs
@@ -1,8 +1,12 @@
-using App.ControlPanel.Engine;
+﻿using App.ControlPanel.Engine;
 using DataUtils;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
 using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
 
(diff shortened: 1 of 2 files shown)


sambetts:
## Review round 1 applied (gpt-5.6-sol, independent review)

The reviewer found six issues, one a blocker. All are now fixed; the suite went from **21 tests to 43, 0 skipped**.

### Blocker: only one direction of compatibility was tested

The golden vectors prove **old -> new** (the current code reads files written by the shipping build). Nothing proved **new -> old**. As the reviewer put it, an implementation that kept legacy decryption as a fallback while *writing* a new format would leave every test green and still produce config files the shipping installer cannot open.

Added `LegacyFormatOracle` — an independent reimplementation of the original format using only BCL primitives (`Rfc2898DeriveBytes` + `RijndaelManaged` at a 256-bit block), which deliberately never calls `StringCipher`. Fresh `Encrypt` output must be readable by it. The oracle is itself validated against the stored golden vectors, so it cannot silently drift away from ground truth.

### The oracle nearly shipped silently disabled

First attempt guarded it with `#if NETFRAMEWORK`. Legacy non-SDK csproj files **do not define that symbol**, so all nine oracle tests reported `Skipped` — the blocker fix would have been inert. Replaced with a runtime capability probe: try to construct a 256-bit-block Rijndael and see. That is also the precise question that matters, and it self-adapts if the project is later multi-targeted.

### Other findings, all fixed

| # | Finding | Fix |
|---|---|---|
| 3 | No non-empty plaintext was block-aligned, so an implementation omitting the extra PKCS7 block for aligned input would go undetected | Added golden vectors at exactly 32 and 64 UTF8 bytes, captured from the shipping build, plus an explicit full-padding-block assertion |
| 4 | Wrong-password tests tolerated garbage; config-level test never asserted `DecryptedOk` | Both now assert exactly (`ThrowsException<CryptographicException>`, `IsFalse(DecryptedOk)`) |
| 5 | Random-IV test compared whole ciphertexts, so it passed if only the salt varied | Samples 8 encryptions and compares the 32-byte salt and 32-byte IV fields independently |
| 6 | Plaintext-leak check used substring matching against a secret containing `"` and `\`, which JSON escapes — a real leak would not have matched | Walks the parsed document and compares decoded string values |

### One reviewer correction to me

The reviewer stated the file had no UTF-8 BOM. I had asserted it did. **They were right** — an `edit` after my original BOM write had stripped it. It compiled anyway because Roslyn defaults to UTF-8, but the BOM is now restored.

### Result

```
Passed!  - Failed: 0, Passed: 43, Skipped: 0, Total: 43 - Tests.UnitTests.dll (net48)
```

Verified against **both** implementations: the original `RijndaelManaged` one on this branch, and the BouncyCastle one in #523. That the legacy oracle can read BouncyCastle's output is the evidence that #523 does not strand config files.

sambetts:
## Review round 2 applied (gpt-5.6-terra) — and it found a real gap in round 1's fix

Round 2 was run specifically because round 1's fixes were ~276 lines of new, unreviewed code. That turned out to be justified.

**Finding (should-fix, 10/10):** every fresh encryption in the new-to-old test used the same ASCII passphrase. `GoldenUnicodePass` was only ever used to *decrypt* a stored vector.

That is a genuine hole. The passphrase is UTF8-encoded inside the KDF, so a writer that changed that conversion — precisely the sort of thing a crypto-library migration gets wrong — would still round-trip in process, still read every stored vector, and still write config files the shipping installer could not open. Only exercising a non-ASCII passphrase on the **write** path catches it.

Fixed: the new-to-old and round-trip tests are now data-driven over `(secret, passphrase)` pairs, including two non-ASCII-passphrase cases. **47 tests, 0 skipped.**

This matters directly for #523: it is now proven that BouncyCastle's `Encoding.UTF8.GetBytes(passPhrase)` inside `Pkcs5S2ParametersGenerator` is byte-equivalent to `Rfc2898DeriveBytes(string, ...)` for non-ASCII passphrases, by having an independent .NET Framework reader decode BouncyCastle's output:

```
Passed Encrypt_OutputIsReadableByAnIndependentLegacyDecoder ("SecretValue-123","Κωδικός-Πρόσβασης")
Passed Encrypt_OutputIsReadableByAnIndependentLegacyDecoder ("Καλημέρα κόσμε","Κωδικός-Πρόσβασης")
```

**Finding (nit, 10/10):** `GoldenVectors_BlockAlignedPlainText_CarriesAFullPaddingBlock` is arithmetic over constants and never calls `StringCipher`, so it would stay green if the writer stopped appending the full padding block. Now documented honestly as fixture validation; writer behaviour for those same aligned inputs is covered by the new-to-old test. The reviewer also independently verified the vector arithmetic (32 bytes -> 128 raw, 64 bytes -> 160 raw), which matches what the shipping build produced.

No other issues found in round 1's fixes.

Issue #509, opened 2026-09-12
CloudInstallEngine does not compile on .NET 10: System.Linq.AsyncEnumerable ambiguity (11 sites)


sambetts:
Part of #519 (.NET 10 parallel branch epic).

---

## Problem

`Common/CloudInstallEngine` does not compile for `net10.0`. **11 sites**, all `CS0121`:

```
error CS0121: The call is ambiguous between the following methods or properties:
'System.Linq.AsyncEnumerable.Where<TSource>(IAsyncEnumerable<TSource>, Func<TSource, bool>)' and
'System.Linq.Enumerable.Where<TSource>(IEnumerable<TSource>, Func<TSource, bool>)'
```

## Cause

.NET 10 added **`System.Linq.AsyncEnumerable`** to the base class libraries (previously only available via the `System.Linq.Async` NuGet package).

Azure Resource Manager collection types — `AppServicePlanCollection`, `RedisCollection`, `SqlFirewallRuleCollection` etc. — implement **both** `IEnumerable<T>` and `IAsyncEnumerable<T>`. Calling `.Where(...)` on one was previously unambiguous because only `Enumerable.Where` was in scope. On .NET 10 both extension methods apply, and the compiler cannot choose.

Example — `Azure/InstallTasks/AppServicePlanTask.cs:29`:

```csharp
var plan = Container.GetAppServicePlans().Where(p => p.Data.Name == _config.ResourceName).SingleOrDefault();
```

This is a **.NET 10 source-breaking change**, not a bug in our code. It only bites because the Azure SDK collections are dual-interface.

## Affected sites

| File | Line |
|---|---|
| `Azure/InstallTasks/AppServicePlanTask.cs` | 29 |
| `Azure/InstallTasks/AppServiceWebsiteTask.cs` | 40 |
| `Azure/InstallTasks/KeyVaultFirewallConfigTask.cs` | 114 |
| `Azure/InstallTasks/KeyVaultTasks.cs` | 212 |
| `Azure/InstallTasks/LogAnalyticsInstallTask.cs` | 29 |
| `Azure/InstallTasks/RedisInstallTask.cs` | 82, 151 |
| `Azure/InstallTasks/ServiceBusNamespaceInstallTask.cs` | 84 |
| `Azure/InstallTasks/SqlServerFirewallConfigTask.cs` | 136, 142 |
| `Azure/InstallTasks/TextAnalyticsInstallTask.cs` | 29 |

## Fix

[...]

PR #522, opened 2026-09-12
Disambiguate LINQ calls on Azure ARM collections

sambetts:
## Scope

Adds `.AsEnumerable()` at 11 call sites in `Common/CloudInstallEngine/Azure/InstallTasks`. **9 files, 11 insertions, 11 deletions.** No behaviour change on .NET Framework 4.8.

## Why

Azure Resource Manager collection types — `AppServicePlanCollection`, `WebSiteCollection`, `RedisCollection`, `SqlFirewallRuleCollection`, `ServiceBusNamespaceCollection` and friends — implement **both** `IEnumerable<T>` and `IAsyncEnumerable<T>`.

Calling `.Where(...)` directly on one resolves unambiguously today only because `System.Linq.Enumerable` is the only candidate in scope. **.NET 10 adds `System.Linq.AsyncEnumerable` to the BCL**, which makes both extension methods applicable and the call ambiguous:

```
error CS0121: The call is ambiguous between the following methods or properties:
'System.Linq.AsyncEnumerable.Where<TSource>(IAsyncEnumerable<TSource>, Func<TSource, bool>)' and
'System.Linq.Enumerable.Where<TSource>(IEnumerable<TSource>, Func<TSource, bool>)'
```

This is a .NET 10 **source-breaking change**, not a defect in our code — it only bites because the Azure SDK collections are dual-interface.

## Why this targets `dev` rather than a .NET 10 branch

`.AsEnumerable()` exists on .NET Framework 4.8, so this compiles identically on the current target. It pins the synchronous overload these sites already resolve to — enumeration semantics, ordering and the number of ARM round-trips are all unchanged.

Landing it here rather than on a port branch keeps the eventual .NET 10 diff smaller, and makes the existing intent explicit at each call site regardless of framework.

## Call sites

| File | Line | Receiver |
|---|---|---|
| `AppServicePlanTask.cs` | 29 | `Container.GetAppServicePlans()` |
| `AppServiceWebsiteTask.cs` | 40 | `Container.GetWebSites()` |
| `KeyVaultFirewallConfigTask.cs` | 114 | `Container.GetWebSites()` |
| `KeyVaultTasks.cs` | 212 | `Container.GetWebSites()` |
| `LogAnalyticsInstallTask.cs` | 28–29 | `Container.GetOperationalInsightsWorkspaces()` |
| `RedisInstallTask.cs` | 82 | `allLegacy` |
| `RedisInstallTask.cs` | 151 | `allClusters` |
| `ServiceBusNamespaceInstallTask.cs` | 84 | `allNSs` |
| `SqlServerFirewallConfigTask.cs` | 135–136 | `server.GetSqlFirewallRules()` (`.Select`) |
| `SqlServerFirewallConfigTask.cs` | 142 | `server.GetSqlFirewallRules()` (`.Where`) |
| `TextAnalyticsInstallTask.cs` | 29 | `Container.GetCognitiveServicesAccounts()` |

## Verification

| Leg | Before | After |
|---|---|---|
| `netstandard2.0` (what ships today) | clean | **0 warnings, 0 errors** |
| `net10.0` (temporary multi-target, reverted before commit) | 22 × `CS0121` | **Build succeeded, 0 errors** |

The `netstandard2.0` leg staying at 0/0 is the evidence that this is safe for the current product.

## Deliberately not in scope

`.Where(...).SingleOrDefault()` on an ARM collection enumerates client-side; several of these could arguably be server-side `Get`/`Exists` calls instead. That would be a behaviour change and is kept out so this PR stays trivially reviewable — every hunk is the same one-token insertion.

## Reviewer note

The change is purely mechanical. The thing worth checking is that `.AsEnumerable()` was inserted **before** `.Where`/`.Select` and not after, so the synchronous overload is selected on the collection rather than on an already-projected sequence.

Addresses #509.

### DIFF
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/ConfigureAzureComponentsTasks.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/ConfigureAzureComponentsTasks.cs
@@ -367,7 +367,7 @@ namespace App.ControlPanel.Engine
                 {
                     var rules = sqlServer.GetSqlFirewallRules();
 
-                    var existing = rules
+                    var existing = rules.AsEnumerable()
                         .Where(r => r.Data.Name == AzurePaaSInstallJob.INSTALLER_FIREWALL_RULE_NAME)
                         .Select(r => new SqlFirewallRuleRange(r.Data.Name, r.Data.StartIPAddress, r.Data.EndIPAddress))
                         .SingleOrDefault();
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/JobTasks/AutomationAccountTask.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/JobTasks/AutomationAccountTask.cs
@@ -43,7 +43,7 @@ namespace App.ControlPanel.Engine.InstallerTasks.Tasks
         public override async Task<AutomationAccountResource> ExecuteTaskReturnResult(object contextArg)
         {
             // Get/create app-service with plan
-            var automationAccount = Container.GetAutomationAccounts().Where(s => s.Data.Name == _config.ResourceName).SingleOrDefault();
+            var automationAccount = Container.GetAutomationAccounts().AsEnumerable().Where(s => s.Data.Name == _config.ResourceName).SingleOrDefault();
             if (automationAccount == null)
             {
                 var newAutomationAccountInfo = new AutomationAccountCreateOrUpdateContent()
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/JobTasks/HybridWorkerGroupTask.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/JobTasks/HybridWorkerGroupTask.cs
@@ -44,7 +44,7 @@ namespace App.ControlPanel.Engine.InstallerTasks.Tasks
             try
             {
                 // Find automation account
-                var automationAccount = Container.GetAutomationAccounts().Where(s => s.Data.Name == automationAccountName).SingleOrDefault();
+                var automationAccount = Container.GetAutomationAccounts().AsEnumerable().Where(s => s.Data.Name == automationAccountName).SingleOrDefault();
                 if (automationAccount == null)
                 {
                     _logger.LogError($"Automation account '{automationAccountName}' not found. Cannot create hybrid worker group.");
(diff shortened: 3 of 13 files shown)


sambetts:
## Review round 1 applied (gemini-3.8-flash, independent review)

The reviewer confirmed the original 11 sites are correct, correctly positioned and behaviour-neutral — and then found **8 more at-risk sites I had missed**, because I only searched `CloudInstallEngine`. `App.ControlPanel.Engine` is the other project referencing the Azure ARM SDK.

All 8 verified against the source before changing anything, and now fixed in commit `12655156`:

| File | Line | Collection type |
|---|---|---|
| `ConfigureAzureComponentsTasks.cs` | 371 | `SqlFirewallRuleCollection` |
| `InstallerTasks/JobTasks/AutomationAccountTask.cs` | 46 | `AutomationAccountCollection` |
| `InstallerTasks/JobTasks/HybridWorkerGroupTask.cs` | 47 | `AutomationAccountCollection` |
| `SolutionInstallVerifier.cs` | 148 | `SqlServerCollection` |
| `SolutionInstallVerifier.cs` | 276 | `AppServicePlanCollection` |
| `SolutionInstallVerifier.cs` | 357 | `KeyVaultCollection` |
| `SolutionInstallVerifier.cs` | 709 | `RedisCollection` |
| `SolutionInstallVerifier.cs` | 722 | `RedisEnterpriseClusterCollection` |

`App.ControlPanel.Engine` builds clean with no new warnings.

### Useful clarification from the review: which operators are actually at risk

Only operators that have an `IAsyncEnumerable` counterpart **of the same name** are affected — `Where`, `Select`, `SelectMany`, `OrderBy`, `Take`, `Skip`, `Distinct`, `Concat`, `Cast`, `OfType` and similar streaming operators.

Terminal operators are **not** at risk: `System.Linq.AsyncEnumerable` spells them `FirstOrDefaultAsync`, `SingleOrDefaultAsync`, `ToListAsync`, `AnyAsync`, `CountAsync`. So a bare `SingleOrDefault()` or `ToList()` directly on an ARM collection stays unambiguous, and those sites are deliberately left alone rather than churned.

The reviewer also verified the three local-variable receivers (`allLegacy`, `allClusters`, `allNSs`) are genuine `ArmCollection` subtypes rather than already-materialised lists, so those three sites were correctly diagnosed and not no-ops.

### Also confirmed not at risk

- `AzureInstallJob.cs:43` — `.ToList()` materialises to `List<T>` first, so the following `.Where` is safe
- `GenericArmSingleResourceInstallTask.cs:114` — `Azure.Pageable<T>` implements only `IEnumerable<T>`
- `KeyVaultTasks.cs:238` — `Pageable<T>` plus a non-ambiguous `FirstOrDefault`

**Scope is now 13 files, 19 call sites.**

Issue #512, opened 2026-09-12
Multi-target the shared class libraries to net48;net10.0


sambetts:
Part of #519 (.NET 10 parallel branch epic).

---

## Goal

Multi-target the shared class libraries to **`net48;net10.0`** so the same source builds for both runtimes. This is the first change that actually lives on the `net10` branch.

Depends on the SDK-style conversion and the `System.Linq.AsyncEnumerable` fix landing on `main` first.

## Order (bottom-up)

1. `Common/DataUtils` — ✅ **already builds clean for `net10.0`** (0 errors)
2. `Common/UsageReporting` — ✅ **already builds clean for `net10.0`** (0 errors, 0 warnings)
3. `Common/CloudInstallEngine` — blocked only by the 11 `CS0121` sites
4. `Common/Entities` — the EF6 model; needs the `DbConfiguration` provider registration
5. `WebJob.*.Engine` libraries
6. `App.ControlPanel.Engine`

### `Common/DataUtils` — obsolete crypto (9 warnings, 0 errors)

`StringCipher.cs` uses APIs that are `[Obsolete]` on modern .NET:

| API | Warning |
|---|---|
| `RijndaelManaged` | `SYSLIB0022` — use `Aes` |
| `Rfc2898DeriveBytes(string, byte[], int)` | `SYSLIB0060` — use static `Pbkdf2` |
| `RNGCryptoServiceProvider` | `SYSLIB0023` — use `RandomNumberGenerator` |

> ⚠️ **Do not "modernise" these casually.** `StringCipher` encrypts secrets in saved installer configuration files. `RijndaelManaged` with a non-default block size is **not** interchangeable with `Aes`, and changing the KDF iteration count or algorithm will make existing saved configs undecryptable. Any change here needs a round-trip compatibility test that decrypts a config produced by the current shipping code. Suppress the warnings for now and treat re-keying as a separate, deliberate piece of work.

### `Common/DataUtils` — DPAPI is Windows-only (`CA1416`)

`StringUtils.cs:451,459` use `ProtectedData.Protect` / `Unprotect` with `DataProtectionScope.CurrentUser`. On .NET 10 these are annotated Windows-only.

[...]

PR #523, opened 2026-09-12
Make installer config encryption work on modern .NET without changing the format

sambetts:
## Stacked on #521

Base is `sambetts/fix-520-cipher-compat` (PR #521, the golden-vector tests), so this diff shows **only the implementation change**. GitHub will retarget this to `dev` automatically when #521 merges.

The tests in #521 are what validate this PR — they were written first, deliberately, so the fix had something to prove itself against.

## Problem

`StringCipher` used `RijndaelManaged` with `BlockSize = 256`. That is Rijndael but **not AES** — AES is Rijndael fixed to a 128-bit block. .NET Framework's `RijndaelManaged` supports the larger block; on .NET Core / .NET 5+ the type is implemented over AES and throws at runtime:

```
System.PlatformNotSupportedException: BlockSize must be 128 in this implementation.
```

It **compiles cleanly** (only `SYSLIB0022` obsolete warnings), so nothing catches it until execution. On a ported installer the effect is that it can neither decrypt an existing configuration file nor write a new one — it fails at the first action an admin takes, on the two secrets that matter most (SQL admin password, app registration client secret).

## Change

.NET has no in-box 256-bit-block Rijndael, so the primitive comes from BouncyCastle's `RijndaelEngine`, which implements the full Rijndael specification.

**The on-disk format is unchanged in every respect:**

| | Value |
|---|---|
| Layout | `[32 bytes salt][32 bytes IV][ciphertext]`, base64 |
| KDF | PBKDF2-HMAC-SHA1, 1000 iterations, 256-bit key |
| Cipher | Rijndael-256 / CBC / PKCS7 |

BouncyCastle is used on **every** target framework rather than `#if`-ing between old and new code, so byte-compatibility between builds is structural rather than something two implementations have to keep agreeing on.

## Verification

**All 21 `StringCipherCompatibilityTests` pass**, including the six golden vectors captured from the shipping .NET Framework 4.8.9310.0 build — so configuration files already in the field still open.

The same implementation was then compiled for `net10.0` and run against those same vectors:

```
Runtime: .NET 10.0.11
  Ascii    OK
  Greek    OK
  Empty    OK
  roundtrip OK
RESULT: ALL PASSED
```

Previously that produced `PlatformNotSupportedException` on every case. This is the actual objective of #520 — .NET 10 can now read secrets written by the .NET Framework build.

### 1. Exception translation is load-bearing

BouncyCastle signals a bad key with `InvalidCipherTextException`, **not** `CryptographicException`. `SolutionInstallConfig.LoadFromJson` catches `CryptographicException` specifically, to report "wrong password" (`DecryptedOk = false`) instead of crashing. Without translation, a wrong password would become an unhandled exception.

The BouncyCastle exceptions are therefore translated at the boundary, and a malformed base64 payload is translated the same way since it is the same outcome from a caller's point of view.

`LoadFromJson_WithWrongPassword_ReportsFailureRatherThanGarbage` covers this and passes.

I deliberately did **not** widen the `catch` in `LoadFromJson` instead. Translating at the source preserves the existing contract without a broad catch that could mask unrelated failures.

### 2. `PrivateAssets=compile` on the package reference

BouncyCastle's root namespace is `Org`, which collides with the solution's own `Org` **entity type** in consuming projects. Adding the package plainly breaks the build:

```
error CS0118: 'Org' is a namespace but is used like a type
  Tests.UnitTests\DataCleanupTests.cs(552,27)
  Tests.UnitTests\EntityTests.cs(695,17), (698,29)
```

Only `StringCipher`'s implementation needs the package, so its compile-time asset is marked `<PrivateAssets>compile</PrivateAssets>` — consumers still get `BouncyCastle.Cryptography.dll` copied to their output (verified present in the test output folder) but do not see the namespace.

This is worth a careful look: if a future change needs BouncyCastle types in another project's *source*, that project must reference the package directly and will hit the same `Org` collision.

## Risk

This is security-critical code on the path that protects customer secrets. The mitigations are that the format is unchanged, the golden vectors prove field compatibility in both directions, and the change is implementation-only — no public API, no call site, and no config-file content changes.

Dependency added: `BouncyCastle.Cryptography` 2.7.0 (widely used, actively maintained, MIT).

## Not in scope

- Re-keying or modernising the KDF/cipher. The 1000-iteration PBKDF2-SHA1 is weak by modern standards, but changing it is a **format** change that would strand existing config files. That needs a versioned format and a migration path — worth a separate issue, and explicitly not this PR.
- `ProtectedData`/DPAPI Windows-only warnings in `DataUtils` (tracked in #512).

Addresses #520.

### DIFF
--- a/src/AnalyticsEngine/Common/DataUtils/DataUtils.csproj
+++ b/src/AnalyticsEngine/Common/DataUtils/DataUtils.csproj
@@ -21,6 +21,13 @@
     <PackageReference Include="Azure.Identity" Version="1.21.0" />
     <PackageReference Include="Azure.Security.KeyVault.Certificates" Version="4.9.0" />
     <PackageReference Include="BCrypt.Net-Next" Version="4.0.3" />
+    <PackageReference Include="BouncyCastle.Cryptography" Version="2.7.0">
+      <!-- BouncyCastle's root namespace is "Org", which collides with the solution's own Org entity
+           type in consuming projects (CS0118: 'Org' is a namespace but is used like a type).
+           Only StringCipher's implementation needs it, so keep the compile-time asset private while
+           still flowing the runtime assembly to consumers' output folders. -->
+      <PrivateAssets>compile</PrivateAssets>
+    </PackageReference>
     <PackageReference Include="EntityFramework" Version="6.5.2" />
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.22.0" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="10.0.11" />
--- a/src/AnalyticsEngine/Common/DataUtils/StringCipher.cs
+++ b/src/AnalyticsEngine/Common/DataUtils/StringCipher.cs
@@ -1,104 +1,157 @@
 ﻿using System;
-using System.IO;
 using System.Linq;
 using System.Security.Cryptography;
 using System.Text;
+using Org.BouncyCastle.Crypto;
+using Org.BouncyCastle.Crypto.Digests;
+using Org.BouncyCastle.Crypto.Engines;
+using Org.BouncyCastle.Crypto.Generators;
+using Org.BouncyCastle.Crypto.Modes;
+using Org.BouncyCastle.Crypto.Paddings;
+using Org.BouncyCastle.Crypto.Parameters;
+using Org.BouncyCastle.Security;
 
 namespace DataUtils
 {
-
+    /// <summary>
+    /// Encrypts the secrets persisted in installer configuration files - the SQL admin password and
+    /// the app registration client secrets.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <b>The on-disk format is fixed and must never change.</b> A configuration file has to open with
+    /// the same password that saved it, whichever build of the installer is opening it. Every format
+    /// detail below is pinned by <c>StringCipherCompatibilityTests</c> against golden vectors captured
+    /// from a shipping build; if those fail, the change has broken every config file in the field.
+    /// </para>
+    /// <para>
+    /// Format: <c>[32 bytes salt][32 bytes IV][ciphertext]</c>, base64-encoded.
+    /// Key derivation: PBKDF2-HMAC-SHA1, 1000 iterations, 256-bit key.
+    /// Cipher: <b>Rijndael with a 256-bit block</b>, CBC, PKCS7.
+    /// </para>
+    /// <para>
+    /// Note that a 256-bit block means this is Rijndael but <b>not AES</b> - AES is Rijndael fixed to a
+    /// 128-bit block. The implementation previously used <c>RijndaelManaged</c>, which supports the
+    /// larger block only on .NET Framework. On .NET Core / .NET 5+ that type is implemented over AES and
+    /// throws <c>PlatformNotSupportedException: BlockSize must be 128 in this implementation.</c>, so the
+    /// installer could neither read nor write a config file once ported. Since .NET has no in-box
+    /// 256-bit-block Rijndael, the primitive comes from BouncyCastle's <see cref="RijndaelEngine"/>,
+    /// which implements the full Rijndael specification and behaves identically on both runtimes.
+    /// </para>
+    /// <para>
+    /// BouncyCastle is used on every target framework rather than conditionally, so byte-compatibility
+    /// between builds is structural rather than something two implementations have to keep agreeing on.
+    /// </para>
+    /// </remarks>
     public static class StringCipher
     {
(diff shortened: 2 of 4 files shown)


sambetts:
## Packaging verified against real CI artifacts

The `<PrivateAssets>compile</PrivateAssets>` trick is the part of this PR most likely to fail *silently in production*: if the compile asset is private but the runtime assembly does not reach a package, the installer throws `FileNotFoundException` in front of a customer, and a green build would not reveal it.

So rather than reasoning about it, I downloaded the artifacts produced by this PR's own CI run (`34686451631`) and looked inside every shipped package:

| Package | `DataUtils.dll` | `BouncyCastle.Cryptography.dll` |
|---|---|---|
| `ControlPanelApp.zip` | yes | **yes** (4.9 MB, alongside `AnalyticsInstaller.exe`) |
| `Office365ActivityImporter.zip` | yes | **yes** |
| `AppInsightsImporter.zip` | yes | **yes** |
| `Website.zip` | yes | **yes** |

So the runtime asset flows correctly to every consumer, including the transitive ones, through the normal packaging path. No `FileNotFoundException` risk.

**Note for release notes:** this adds roughly 4.9 MB to each download. Worth a line for admins on metered or air-gapped transfers, though it is small relative to the existing package sizes.

## Bidirectional compatibility now proven by automated test

Following the review of #521, the test suite grew from 21 to **43 tests** and now covers both directions rather than one:

- **old -> new** — golden vectors captured from the shipping .NET Framework build
- **new -> old** — `LegacyFormatOracle`, an independent BCL-only reimplementation of the original format, must be able to read what this implementation writes

That second direction is what guards this PR specifically. Result against the BouncyCastle implementation:

```
Passed!  - Failed: 0, Passed: 43, Skipped: 0, Total: 43 - Tests.UnitTests.dll (net48)
```

Including all eight `Encrypt_OutputIsReadableByAnIndependentLegacyDecoder` cases — so ciphertext written by BouncyCastle is decodable by a plain .NET Framework `RijndaelManaged` reader. A config saved by the ported installer still opens in the shipping one.

sambetts:
## Independent review round 1 (claude-opus-4.7): **no significant issues found**

Reviewed adversarially against eight specific concerns. Summary of what was independently verified rather than reasoned about:

**PBKDF2 equivalence is exact.** `Pkcs5S2ParametersGenerator(new Sha1Digest())` seeded with the UTF-8 bytes of the passphrase, 1000 iterations, 256-bit output is byte-identical to `new Rfc2898DeriveBytes(string, byte[], int).GetBytes(32)`. The `"AES"` string in `GenerateDerivedParameters` only selects the wrapper subtype and yields the same bytes as `"RIJNDAEL"` — verified by loading BouncyCastle 2.7.0 and comparing outputs directly. Unknown names throw rather than silently taking a wrong path.

**No input divergence found.** The reviewer ran the new implementation against the shipping build's own `RijndaelManaged` + `Rfc2898DeriveBytes` (hosted on .NET Framework 4.8.9310.0) across empty plaintext, exact block boundaries at 32 and 64 bytes, 500 and 10,000 characters, embedded null bytes, newlines, non-Latin plaintext and passphrase, **non-BMP UTF-8 (emoji)**, and a 1024-character passphrase. All nine bidirectional round-trips match.

**The old single-`Read` pattern was checked empirically.** `cryptoStream.Read(buf, 0, buf.Length)` on 4.8.9310.0 returned everything in one call for the tested sizes, so the shipping build never stored a truncated secret and the new implementation returning "more" data is not a scenario existing files can reach.

**Exception hierarchy confirmed by reflection** against BouncyCastle 2.7.0: `InvalidCipherTextException`, `DataLengthException` and `OutputLengthException` all derive from `CryptoException`, so the translation covers padding failures, non-block-multiple ciphertext and short-output paths. The only non-`CryptoException` path (`ArgumentException` on key/IV size) is unreachable given the length guard.

**An unintended improvement, worth knowing:** malformed input that the old code leaked as `IndexOutOfRangeException` or `FormatException` now surfaces as `CryptographicException` — which `SolutionInstallConfig.LoadFromJson` actually catches. So a corrupt or truncated config file now degrades to "could not decrypt" instead of crashing the installer. That is a bug fix, not a compatibility break.

**Null-guard behaviour is unchanged in type**, only `ParamName` differs, and both call sites gate on `!string.IsNullOrEmpty` anyway.

**`PrivateAssets=compile` verified end to end** — independently reaching the same conclusion as the artifact check above, and additionally confirming the `lib/net461` variant is the one selected, that no binding redirect is needed (single identity 2.0.0.0), and that `ci.yml` filters nothing out of the zip.

**Stateless and thread-safe**: no `IDisposable` BouncyCastle types are leaked, nothing shares state. Wrong-password detection remains unauthenticated CBC+PKCS7 — a property of the format, not of this port.

No changes required as a result of this review.

Issue #511, opened 2026-09-12
EF6 on .NET 10: confirmed working by spike; adopt Microsoft.EntityFramework.SqlServer


sambetts:
Part of #519 (.NET 10 parallel branch epic).

---

## Summary

**Confirmed by spike: EF6 runs on .NET 10 essentially unchanged — model, `Migrations/` folder, `.resx` snapshots, `DbMigrator` and interceptors all work.** This was the main open question for the `net10` branch and the answer is yes.

The one change worth making is **swapping the SQL provider** from the deprecated `System.Data.SqlClient` to `Microsoft.Data.SqlClient` via the official **`Microsoft.EntityFramework.SqlServer`** package. That change is **net48-safe and should land on `main` first** — it is valuable on its own, independent of .NET 10.

## Evidence

Two console spikes, both `net10.0`, run against LocalDB on **.NET 10.0.11 / SDK 10.0.400**.

`EntityFramework` 6.5.2 ships a `netstandard2.1` asset (`lib/net40`, `lib/net45`, `lib/netstandard2.1`), which is what makes this possible — NuGet resolves it for `net10.0` without complaint.

### Spike A — current shape (EF6 + `System.Data.SqlClient` 4.9.1)

```
Runtime           : .NET 10.0.11
EF informational  : 6.5.2
EF target fwk     : .NETStandard,Version=v2.1
SqlClient assembly: System.Data.SqlClient 4.6.1.6

[1] DbMigrator.Update()                      OK  -> 202609120714016_AutomaticMigration
[2] Unicode round-trip 'Καλημέρα κόσμε'      OK
[3] LINQ->SQL translated                     OK
[4] __MigrationHistory rows: 1               OK
[5] Widgets.Name column type: nvarchar       OK
[6] IDbConnectionInterceptor fired 25x       OK
RESULT: ALL CHECKS PASSED
```

### Spike B — recommended shape (EF6 + `Microsoft.EntityFramework.SqlServer` 6.5.2 + `Microsoft.Data.SqlClient` 7.0)

```
EF                : 6.5.2 (.NETStandard,Version=v2.1)
SqlClient         : Microsoft.Data.SqlClient 7.0.0.0
EF6 provider      : Microsoft.EntityFramework.SqlServer 6.0.0.0

[1] DbMigrator.Update()                      OK  -> 202609120725258_AutomaticMigration
[2] Unicode round-trip 'Καλημέρα κόσμε'      OK
[...]

PR #525, opened 2026-09-12
Switch EF SQL provider to Microsoft.Data.SqlClient

sambetts:
## Summary
Addresses #511.

This switches the EF6 SQL Server provider path from the deprecated `System.Data.SqlClient` provider to the official Microsoft.Data-backed provider:

- Replaced C# `SqlConnection`/`SqlCommand`/`SqlException` usages with `Microsoft.Data.SqlClient`.
- Added `Microsoft.Data.SqlClient` 6.1.4 and `Microsoft.EntityFramework.SqlServer` 6.5.2 package references to the .NET Framework entry points/libraries that need them; `Common.DataUtils` keeps only `Microsoft.Data.SqlClient` because the EF provider package has no netstandard2.0 asset.
- Updated EF provider configuration to `Microsoft.Data.SqlClient`, `MicrosoftSqlProviderServices`, `MicrosoftSqlDbConfiguration`, `MicrosoftSqlAzureExecutionStrategy`, and the Microsoft SQL connection factories.
- Updated provider factories and binding redirects in the application templates/configs.
- Removed all tracked `System.Data.SqlClient` source/config/project references.
- Added `TrustServerCertificate=True` only for LocalDB/test/demo connection strings and LocalDB-only scratch database builders.

## Behavioural risk

`Microsoft.Data.SqlClient` defaults to `Encrypt=True`; the old provider defaulted to `Encrypt=False`. Azure SQL deployments should generally benefit from the stricter default, but non-Azure/dev SQL servers with untrusted certificates can fail during connection open unless their connection strings explicitly opt into a trusted certificate chain or a deliberate `TrustServerCertificate=True` dev/test override.

I intentionally did **not** add `TrustServerCertificate=True` to customer deployment connection-string builders or installer production paths because that would weaken certificate validation. Reviewers should check any non-Azure SQL deployment guidance before merge.

## Not changed

- Kept the existing `AzureSqlTokenAuth` / `AzureSqlAccessTokenInterceptor` compatibility workaround. `Microsoft.Data.SqlClient` can handle modern `Authentication=Active Directory ...` keywords natively, so removing/simplifying that fallback is a follow-up.
- No installer saved-config schema change; `CONFIG_VERSION` was not bumped.

## Validation

- `MSBuild.exe src\AnalyticsEngine\O365 Advanced Analytics Engine.sln -p:Configuration=Release -p:Platform="Any CPU" -m -v:minimal` succeeded. It still emits the pre-existing `System.Text.Json` binding-remap warnings.
- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests\Tests.UnitTests.csproj -p:Configuration=Debug -p:Platform=anycpu -m -v:minimal` succeeded.
- `vstest.console.exe src\AnalyticsEngine\Tests.UnitTests\bin\Debug\Tests.UnitTests.dll /Platform:x64 /TestCaseFilter:FullyQualifiedName~AzureSqlEntraAuthTests` passed: 68 passed, 0 failed.
- Full Debug `Tests.UnitTests` run completed: 2132 passed, 12 failed, 4 skipped, 2148 total. The failures are existing environment/config-dependent tests against unavailable fake external services or missing local installer test config copy, not SQL-provider failures:
  - `CommentsCognitiveTests`
  - `UsageStatsReporterRealTests`
  - `SharePointSitesUsageLoaderTest`
  - `MessageImportTests`
  - `CallQueueProcessorTest`
  - `SingleTeamSaveToSQLTest`
  - `AllTeamsLoadTest`
  - `ChatMessageLoadCognitiveStatsFromCacheOrAITests`
  - `ConfigFileLoadsWithVNetDefaults`
  - `SolutionInstallConfig_VNetEnabled_SkuUpgradeValidation`
  - `SolutionInstallConfig_SerializationRoundTrip_PreservesVNet`
  - `DeployWithVNet_CreatesVNetAndResources`
- `git grep -n "System.Data.SqlClient" -- '*.cs' '*.config' '*.csproj'` returns no tracked matches.
- Changed config files parse as XML and `git diff --check` passes.

## Reviewer hotspots

- EF provider registration in `Common\Entities\AnalyticsEntitiesContext.cs` and every `<entityFramework>` config block.
- Binding redirects/provider factories in app templates and tracked app configs.
- LocalDB-only `TrustServerCertificate=True` additions versus production/customer connection-string paths.
- The retained `AzureSqlTokenAuth` fallback against the new `Microsoft.Data.SqlClient.SqlConnection` type.

### DIFF
--- a/deploy/Debug Config Files/Unit tests/App.Debug.config
+++ b/deploy/Debug Config Files/Unit tests/App.Debug.config
@@ -27,8 +27,8 @@
 
 	<connectionStrings xdt:Transform="Replace">
 		
-		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework" providerName="System.Data.SqlClient" />
-		<add name="UnitTestingOffice365Services" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=FakeOfficeServicesDB;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework" providerName="System.Data.SqlClient" />
+		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework;TrustServerCertificate=True" providerName="Microsoft.Data.SqlClient" />
+		<add name="UnitTestingOffice365Services" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=FakeOfficeServicesDB;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework;TrustServerCertificate=True" providerName="Microsoft.Data.SqlClient" />
 		<add name="redis" connectionString="" />
 		<add name="ServiceBus" connectionString="" />
 		<add name="Storage" connectionString="UseDevelopmentStorage=true" />
--- a/deploy/Debug Config Files/Web/Web.Debug.config
+++ b/deploy/Debug Config Files/Web/Web.Debug.config
@@ -31,7 +31,7 @@
 	</appSettings>
 
 	<connectionStrings xdt:Transform="Replace">
-		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework" providerName="System.Data.SqlClient" />
+		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework;TrustServerCertificate=True" providerName="Microsoft.Data.SqlClient" />
 		<add name="redis" connectionString="" />
 		<add name="ServiceBus" connectionString="" />
 		<add name="Storage" connectionString="UseDevelopmentStorage=true" />
--- a/deploy/Debug Config Files/Webjobs/App.Debug.config
+++ b/deploy/Debug Config Files/Webjobs/App.Debug.config
@@ -31,7 +31,7 @@
 	</appSettings>
 
 	<connectionStrings xdt:Transform="Replace">
-		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework" providerName="System.Data.SqlClient" />
+		<add name="SPOInsightsEntities" connectionString="Data Source=(localdb)\MSSQLLocalDB;Initial Catalog=UnitTestingSPOInsights;Integrated Security=true;MultipleActiveResultSets=True;App=EntityFramework;TrustServerCertificate=True" providerName="Microsoft.Data.SqlClient" />
 		<add name="redis" connectionString="" />
 		<add name="ServiceBus" connectionString="" />
 		<add name="Storage" connectionString="UseDevelopmentStorage=true" />
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/App.ControlPanel.Engine.csproj
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/App.ControlPanel.Engine.csproj
@@ -217,6 +217,12 @@
     <PackageReference Include="EntityFramework">
       <Version>6.5.2</Version>
     </PackageReference>
+    <PackageReference Include="Microsoft.Data.SqlClient">
+      <Version>6.1.4</Version>
+    </PackageReference>
+    <PackageReference Include="Microsoft.EntityFramework.SqlServer">
+      <Version>6.5.2</Version>
+    </PackageReference>
     <PackageReference Include="Microsoft.ApplicationInsights">
       <Version>2.22.0</Version>
     </PackageReference>
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/BaseInstallProcessClasses.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/BaseInstallProcessClasses.cs
@@ -7,7 +7,7 @@ using DataUtils.Sql;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
-using System.Data.SqlClient;
+using Microsoft.Data.SqlClient;
 using System.Threading.Tasks;
 
 namespace App.ControlPanel.Engine
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlDatabaseUserGrantTask.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlDatabaseUserGrantTask.cs
@@ -4,7 +4,7 @@ using DataUtils.Sql;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
-using System.Data.SqlClient;
+using Microsoft.Data.SqlClient;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlEntraAccessBootstrap.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlEntraAccessBootstrap.cs
@@ -5,7 +5,7 @@ using DataUtils.Sql;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
-using System.Data.SqlClient;
+using Microsoft.Data.SqlClient;
 using System.Threading;
 using System.Threading.Tasks;
 
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlIdentityAccessTask.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/InstallerTasks/SqlIdentityAccessTask.cs
@@ -6,7 +6,7 @@ using DataUtils.Sql;
 using Microsoft.Extensions.Logging;
 using System;
 using System.Collections.Generic;
-using System.Data.SqlClient;
+using Microsoft.Data.SqlClient;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
--- a/src/AnalyticsEngine/App.ControlPanel.Engine/Models/DatabasePaaSInfo.cs
+++ b/src/AnalyticsEngine/App.ControlPanel.Engine/Models/DatabasePaaSInfo.cs
@@ -80,7 +80,7 @@ namespace App.ControlPanel.Engine.Entities
         /// <c>DataUtils.Sql.AzureSqlTokenAuth</c>, which attaches an access token at connection-open time,
         /// and to anything reading the App Service configuration. Deliberately NOT
         /// <c>Authentication=Active Directory Managed Identity</c>: that keyword makes SqlClient acquire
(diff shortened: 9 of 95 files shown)


sambetts:
## Independently verified: the 12 test failures are pre-existing

The PR states the 12 full-suite failures are environment-dependent rather than provider-related. Rather than take that on trust, I re-ran a representative sample on a **baseline worktree off `dev`** carrying the original `System.Data.SqlClient` provider and the original `StringCipher` — i.e. none of this PR's changes.

They fail identically there:

| Test | Baseline result | Cause visible in baseline |
|---|---|---|
| `CommentsCognitiveTests` | Failed | DNS: `The remote name could not be resolved: 'example-cognitive.cognitiveservices.azure.com'` |
| `UsageStatsReporterRealTests` | Failed | Redis RBAC token acquisition against a synthetic tenant GUID |
| `ConfigFileLoadsWithVNetDefaults` | Failed | — |
| `SolutionInstallConfig_VNetEnabled_SkuUpgradeValidation` | Failed | — |
| `SolutionInstallConfig_SerializationRoundTrip_PreservesVNet` | Failed | — |
| `DeployWithVNet_CreatesVNetAndResources` | Failed | — |

So the claim holds: **these are not regressions from the provider swap.** They need real credentials or local installer test config that a fresh worktree does not have.

Worth noting the four `VNet`/`SolutionInstallConfig` ones did not *look* environment-dependent from their names, which is exactly why they were worth checking rather than accepting.

`AzureSqlEntraAuthTests` — the main test for the area this PR touches — passed 68/68.

## Still needed before this leaves draft

The `Encrypt=True` default change is the real risk and is not something tests here can settle: this product is installed against customers' own SQL Servers, not only Azure SQL, and a self-signed or untrusted certificate that worked under `System.Data.SqlClient` will now fail to connect. An independent review is running specifically on that question and on whether any `TrustServerCertificate=True` crept into a customer-facing path rather than a LocalDB fixture.

sambetts:
## Audited the `TrustServerCertificate` additions independently

This is the security-critical claim in the PR, so I checked every added occurrence rather than relying on the description. All nine are confined to test and developer-debug paths, and every one targets `(localdb)\MSSQLLocalDB`:

| File | Lines | Nature |
|---|---|---|
| `deploy/Debug Config Files/Unit tests/App.Debug.config` | 2 | developer debug config |
| `deploy/Debug Config Files/Web/Web.Debug.config` | 1 | developer debug config |
| `deploy/Debug Config Files/Webjobs/App.Debug.config` | 1 | developer debug config |
| `Tests.FakeDataGen/Demo/SqlDemoDatabase.cs` | 1 | demo data generator |
| `Tests.UnitTests/App.Release.config` | 2 | test project config |
| `Tests.UnitTests/DlpApiSqlIntegrationTests.cs` | 2 | test fixture |
| `Tests.UnitTests/LicenceActivitySqlFixture.cs` | 1 | test fixture |
| `Tests.UnitTests/ScratchDatabase.cs` | 2 | test fixture |
| `Tests.UnitTests/StressHarness/ActivityApiDbStressRunner.cs` | 1 | stress harness |

**Nothing was added to a customer-facing connection-string builder** — not in `App.ControlPanel.Engine`, `CloudInstallEngine`, or the shipped `App.Template.config` files. The claim holds.

## The part that is still open

Auditing the diff proves nothing *weakening* was added. It does **not** settle the upgrade question, which is about connection strings this PR never touches:

Customers already have connection strings persisted in installer config files and App Service settings, written when the default was `Encrypt=False`. After upgrading, `Microsoft.Data.SqlClient` will apply `Encrypt=True` to those same unchanged strings. For Azure SQL that is fine and arguably an improvement. For a **customer-managed SQL Server with a self-signed or otherwise untrusted certificate**, the connection will start failing — and the product's own tests reference non-Azure servers (`sql01.contoso.local`), so that configuration is clearly contemplated.

Before this leaves draft it needs an explicit decision on that path: either confirm on-prem SQL is unsupported, or add handling plus a release note telling admins what to change. An independent review is running on exactly this question.

sambetts:
## Review round 1 (grok-4.6) found a **blocker**. Fixed in `31e7beab`.

The reviewer's central finding was not the `Encrypt=True` default I had been worrying about — it was the **provider invariant name**, and it would have broken every existing deployment on upgrade.

### The problem

The installer saves `SPOInsightsEntities` on the App Service with type `SQLAzure` (`ConfigureAzureComponentsTasks.cs:562`). Azure surfaces that as `SQLAZURECONNSTR_*` and hands it to a .NET Framework app with **`providerName="System.Data.SqlClient"`** — the old invariant name — regardless of what the build prefers. Every existing deployment already has that setting, and re-running the installer does not change it because the installer still writes the same type.

After the swap EF knew only `Microsoft.Data.SqlClient`, so the first query after upgrade would fail:

```
No Entity Framework provider found for the ADO.NET provider with invariant name 'System.Data.SqlClient'
```

**The quieter failure mattered more.** If the machine-wide factory resolved instead, the connection would be a `System.Data.SqlClient.SqlConnection` — and `AzureSqlAccessTokenInterceptor` casts to the `Microsoft.Data.SqlClient` type, so it would silently stop attaching Entra tokens. Token-authenticated installs would break with nothing visibly wrong at the provider layer.

The existing tests could not catch any of this: their LocalDB fixtures set `providerName="Microsoft.Data.SqlClient"` explicitly, so they never exercise the name a real deployment supplies.

### The fix

Register the legacy invariant name as an alias for the modern provider, in **both** `SPOInsightsDBConfiguration` and the `entityFramework` config sections (13 config files). Whichever name arrives, the connection is a `Microsoft.Data.SqlClient` one and the Azure retry strategy still applies.

### New regression test, and why the config alias alone is not enough

Added `EfProviderRegistrationTests` (7 tests). Mutation-tested by removing the **code** registrations while leaving the **config** aliases in place:

```
Failed ConnectionCreatedForLegacyInvariantName_IsAMicrosoftDataSqlConnection
  Expected: Microsoft.Data.SqlClient.SqlConnection
  Actual:   System.Data.SqlClient.SqlConnection

Failed ExecutionStrategy_IsRegisteredForBothInvariantNames (legacy)
  Expected: MicrosoftSqlAzureExecutionStrategy
  Actual:   DefaultExecutionStrategy

3 of 7 failed
```

So the config `<provider>` entry is **not** sufficient on its own — it satisfies provider *services* but not the *factory*, which is what decides the connection type and therefore whether the Entra interceptor works at all. The code-based `SetProviderFactory` is doing the load-bearing work. That confirms the reviewer's deeper concern empirically rather than by argument.

Restored: **75 passed, 0 failed** across `EfProviderRegistrationTests` and `AzureSqlEntraAuthTests`.

### On `Encrypt=True`

The reviewer's assessment: for this product it is a non-issue. The installer only ever targets Azure SQL (`ConnectionStringType.SqlAzure`, an ARM SQL server FQDN), whose `*.database.windows.net` certificate chains to a public CA — so enabling validation succeeds. Self-signed on-prem SQL is not an installer path. Their explicit advice, which I agree with: **do not add `TrustServerCertificate=True` to any production path** to paper over this.

I had flagged the on-prem case as needing a decision; the reviewer traced the actual connection-string paths and showed it is not reachable through the installer. That is a better answer than mine.

### Still draft?

The blocker is fixed and covered by tests. Remaining before this leaves draft is a human call on whether any customer is known to point this at a non-Azure SQL Server outside the installer path.

sambetts:
## Review round 2 (gpt-5.6-sol) found that **round 1's fix introduced a worse bug**. Fixed in `00947c9d`.

This is the failure mode the repo's `release-critic` guidance warns about, and it happened exactly as described: the fix for one round became the next round's defect.

### What round 1's fix broke

`SetProviderFactory` registers a **reverse** `IProviderInvariantName` lookup keyed on the factory *instance*, not just the forward name-to-factory one. `MicrosoftSqlDbConfiguration` has already registered the modern name by the time the subclass constructor runs, so the later legacy registration won — and EF began reporting **every** `Microsoft.Data.SqlClient` connection as `"System.Data.SqlClient"`.

Provider-specific services are keyed on the modern name only. `MicrosoftSqlServerMigrationSqlGenerator` therefore stopped resolving, so `DatabaseUpgrader` — which builds a context from a raw connection string and calls `Database.Initialize(true)` — would fail on any upgrade with pending migrations:

```
No MigrationSqlGenerator found for provider 'System.Data.SqlClient'
```

**Ordinary queries kept working.** That is what made it dangerous: it would have surfaced only when a customer upgraded their database, which is the worst possible moment.

Verified before fixing — the three new tests fail against `31e7beab`:

```
Failed ModernFactory_StillReportsTheModernInvariantName
  Expected:<Microsoft.Data.SqlClient>. Actual:<System.Data.SqlClient>.
Failed MigrationSqlGenerator_ResolvesForTheModernProvider
Failed ContextFromRawConnectionString_ResolvesAMigrationSqlGenerator
  EF describes this connection as provider 'System.Data.SqlClient', and no
  MigrationSqlGenerator is registered for that name.
```

### The fix

Replaced the two `Set*` calls with `LegacyProviderNameAliasResolver`, a **forward-only** `IDbDependencyResolver`. It answers *"what services does the legacy name need?"* — provider services, provider factory, migrations SQL generator — and never answers *"what is this factory called?"*, so the modern name stays the provider's single identity.

**10 passed, 0 failed.**

### Why the original seven tests missed it

They only exercised forward lookups (`GetService<DbProviderFactory>("System.Data.SqlClient")`). The regression lived in the reverse direction, which nothing asserted. The three added tests cover it, and they are the ones that fail against the previous commit.

I have kept the `<provider invariantName="System.Data.SqlClient">` config entries: they are harmless, and they cover the case where EF reads provider services straight from configuration.

sambetts:
## Review round 3 (claude-opus-4.7): **clean**

Three rounds, three different models. Rounds 1 and 2 each found a blocker — round 2's being a defect in round 1's fix. Round 3 set out on the assumption that round 2's fix was also broken, and could not break it.

The verification went well beyond reading the diff: the reviewer **disassembled `EntityFramework.dll` 6.5.2 and `Microsoft.EntityFramework.SqlServer.dll` 6.5.2 with ildasm** and traced every EF6 path that resolves a service by provider invariant name.

### Why the deliberate `TableExistenceChecker` omission is safe

I dropped it because the concrete type is not publicly accessible. The reviewer confirmed that and showed why it does not matter: every by-invariant-name lookup in this codebase **except** the initial `DbProviderFactory` resolution is preceded by a *reverse* lookup on the connection's factory instance —

```
InternalContext.get_ProviderName()
  = resolver.GetService<IProviderInvariantName>(DbProviderServices.GetProviderFactory(connection)).Name
```

— which now correctly returns `"Microsoft.Data.SqlClient"`, and `MicrosoftSqlProviderServices`' own inner resolver chain answers `TableExistenceChecker`, `Func<MigrationSqlGenerator>` and `DbSpatialServices` under that name. Also confirmed: `Microsoft.EntityFramework.SqlServer.dll` exports exactly 7 public types and `SqlTableExistenceChecker` is not among them, so the omission was not avoidable anyway.

### Why the resolver is load-bearing and the config entries are not

`AddDependencyResolver` appends to the *Second* chain; the app-config chain is First. `AppConfigDependencyResolver` only ever answers `DbProviderServices`, `IDbConnectionFactory` and `IDatabaseInitializer<T>` — **never `DbProviderFactory`**.

So for the legacy name: the config `<provider>` entry answers `DbProviderServices` first and short-circuits the resolver (both return the same instance, so no conflict), while `DbProviderFactory` is answered *only* by the resolver. That is what stops EF falling back to the machine-wide `DbProviderFactories.GetFactory("System.Data.SqlClient")` and handing back a legacy `SqlConnection` — the exact silent failure that would disable the Entra token interceptor.

In other words the config entries are harmless belt-and-braces; the code resolver is the part doing the work.

### Other points confirmed

- Returning `null` for a non-matching key is the correct contract — `ResolverChain.GetService` filters on `x != null` to mean "try the next resolver". `Enumerable.Empty<object>()` from `GetServices` matches.
- The `Func<MigrationSqlGenerator>` shape is exactly how the modern provider registers the same service internally, per-call instantiation included.
- No path in the product depends on the reverse lookup returning the legacy name. `AzureSqlAccessTokenInterceptor` casts the *connection*, not the name.
- `SetExecutionStrategy` for the legacy name must stay: the execution-strategy key is not run through the reverse lookup, so the App Service-injected name flows in directly.

### Test improvement taken

The reviewer noted `ContextFromRawConnectionString_ResolvesAMigrationSqlGenerator` passed the *modern* name to `DbConnectionInfo`, so it tested the equivalence class rather than literally reproducing the App Service path — safe, but not maximally honest. Changed it to pass the **legacy** name and additionally assert that EF settles on the modern one.

Issue #510, opened 2026-09-12
Convert the 9 remaining legacy .csproj files to SDK-style (staying on net48)


sambetts:
Part of #519 (.NET 10 parallel branch epic).

---

## Goal

Convert the 10 remaining legacy (non-SDK) `.csproj` files to **SDK-style**, while **staying on `net48`**. No framework change, no behaviour change.

This is the single biggest friction-reducer for the `net10` branch: once a project is SDK-style, moving it to .NET 10 is a one-line `TargetFrameworks` edit instead of a project rewrite.

## Current state

Already SDK-style (3):

| Project | TFM |
|---|---|
| `Common/CloudInstallEngine` | `netstandard2.0` |
| `Common/DataUtils` | `netstandard2.0` |
| `Common/UsageReporting` | `netstandard2.0` |

Legacy format, `v4.8` (10):

| Project | Output |
|---|---|
| `Common/Entities` | Library |
| `App.ControlPanel.Engine` | Library |
| `App.ControlPanel.WinForms` | WinExe |
| `WebJob.Office365ActivityImporter` | Exe |
| `WebJob.Office365ActivityImporter.Engine` | Library |
| `WebJob.AppInsightsImporter` | Exe |
| `WebJob.AppInsightsImporter.Engine` | Library |
| `Tests.FakeDataGen` | Exe |
| `Tests.UnitTests` | Library |
| `Web` | Library (**excluded — see below**) |

## Good news

**There is no `packages.config` anywhere in the solution** — every project is already on `PackageReference`. That removes the usual first and most error-prone step entirely. The conversion is therefore mostly about deleting boilerplate rather than migrating dependencies.

## Scope

Convert **9** projects. **Exclude `Web.csproj`**: ASP.NET MVC 5 projects cannot be SDK-style (there is no `System.Web` project SDK, and NuGet's PackageReference migration is explicitly unsupported for ASP.NET projects). The web app gets a new project instead — see the Web migration issue.

## Per-project work

- Replace the project header with `<Project Sdk="Microsoft.NET.Sdk">`
- `<TargetFrameworkVersion>v4.8</TargetFrameworkVersion>` → `<TargetFramework>net48</TargetFramework>`
- Delete the explicit `<Compile Include="..." />` list — but **carefully**: see the migrations caveat below
- Keep `<Reference Include="System.Configuration" />` etc. where still genuinely needed (SDK-style net48 projects do not auto-reference all of the old default set)
[...]

PR #527, opened 2026-09-12
Pin the manifest names of every embedded resource loaded at runtime

sambetts:
## Scope

Adds `Tests.UnitTests/EmbeddedResourceNameTests.cs` — 5 tests, no production code changes. One new test file plus its `<Compile>` entry in the legacy csproj.

## Why

Several features locate their SQL by asking the assembly for an embedded resource, either by exact name or by prefix:

```csharp
// WebJob.Office365ActivityImporter.Engine/ActivityAPI/Copilot/ICopilotStagingWriter.cs:127
_rr.ReadResourceString("WebJob.Office365ActivityImporter.Engine.ActivityAPI.Copilot.SQL.common_upsert_copilot_agents.sql")

// App.ControlPanel.Engine/Utils/DatabaseUpgrader.cs:84
var sqlScriptNames = rr.GetResourceNamesMatchingPathRoot(SqlResourceNameStart);
```

A manifest name is derived from the project's **root namespace and the file's folder path**, so it is silently sensitive to project-file changes — moving a file, renaming a folder, changing `RootNamespace`, or converting a project to SDK style (which *globs* resources instead of listing them).

**Nothing currently catches that.** The code compiles, the build succeeds, and the failure only appears when the feature runs.

### The worst case fails silently rather than loudly

`DatabaseUpgrader` discovers its SQL upgrade scripts by prefix. If the prefix changes it does **not** throw — it finds zero scripts, and the database upgrade quietly does nothing. A customer would see a successful upgrade and an un-upgraded database.

## What is asserted

| Test | Guards |
|---|---|
| `ControlPanelEngine_EmbeddedResourceNames_AreUnchanged` | full name set (4 resources) |
| `ImporterEngine_EmbeddedResourceNames_AreUnchanged` | full name set (15 resources — 13 SQL scripts, a licensing CSV, and `Resources.resources`) |
| `DatabaseUpgrader_StillFindsItsSqlScriptsByPrefix` | the prefix still matches exactly 3 scripts |
| `Entities_StillEmbedsTheDataCleanupScript` | `Common.Entities.Clean Old Data Data.sql` — included from **outside** the project directory (`..\..\..\`), making it the most fragile of the lot, and the spaces are part of the name |
| `Entities_EveryMigrationKeepsItsModelSnapshotResource` | all **58** EF migrations still have their compiled `.resources` entry |

### That last one matters most

The migration `.resources` entries carry the gzipped EDMX model snapshots EF compares the live model against. If those names change, EF cannot find the snapshot for the latest migration, concludes the model has changed, and — with `AutomaticMigrationsEnabled = true` — either auto-migrates or throws `AutomaticDataLossException` on **every context construction**.

It is derived rather than hard-coded: it enumerates types implementing `IMigrationMetadata` and requires a matching `<typeName>.resources`. So it covers migrations added after this PR automatically, with no list to maintain.

## Verification

```
Passed!  - Failed: 0, Passed: 5, Skipped: 0, Total: 5 - Tests.UnitTests.dll (net48)
```

**Mutation-tested.** Removing a single `<EmbeddedResource>` entry from `Entities.csproj`:

```
Failed Entities_EveryMigrationKeepsItsModelSnapshotResource
  Assert.AreEqual failed. Expected:<0>. Actual:<1>.
  1 of 58 migrations have no matching model-snapshot resource.
  Missing: Common.Entities.Migrations.v106.resources
```

Restored, 5/5 green. The other four tests correctly stayed green for that mutation, since it did not affect their assemblies.

## Why now

Written ahead of the SDK-style project conversion in #510, which is the change most likely to trip these — but they guard any accidental rename equally well, and they are worth having regardless of whether that conversion ever happens.

I went looking at #510 and stopped when I found this: the risk is broader than the EF snapshots I had originally documented on that issue. It spans embedded `.sql` scripts across three projects, including the database-upgrade path. Landing the guard first makes that conversion a much safer piece of work.

## Reviewer notes

- `SqlExtensionsPrefix` is deliberately a literal in the test rather than a reference to `DatabaseUpgrader`'s private constant, so the test fails if the constant is changed away from the resources it is meant to match.
- Expected name lists were captured from the currently-built assemblies, not hand-written.
- No database, network or external dependency; runs in under a second.

Relates to #510.

### DIFF
--- a/src/AnalyticsEngine/Tests.UnitTests/EmbeddedResourceNameTests.cs
+++ b/src/AnalyticsEngine/Tests.UnitTests/EmbeddedResourceNameTests.cs
@@ -1,0 +1,169 @@
+﻿using App.ControlPanel.Engine;
+using Common.Entities;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Tests.UnitTests
+{
+    /// <summary>
+    /// Pins the manifest names of every embedded resource the product loads at runtime.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// Several features find their SQL by asking the assembly for an embedded resource, either by exact
+    /// name or by prefix. A resource's manifest name is derived from the project's root namespace and the
+    /// file's folder path, so it is <b>silently</b> sensitive to project-file changes - moving a file,
+    /// renaming a folder, changing <c>RootNamespace</c>, or converting a project to the SDK style, which
+    /// globs resources rather than listing them.
+    /// </para>
+    /// <para>
+    /// Nothing else catches that. The code compiles, the build succeeds, and the failure appears only when
+    /// the feature runs. The worst case is <see cref="DatabaseUpgrader"/>: it discovers its SQL upgrade
+    /// scripts by prefix, so a changed prefix does not throw - it simply finds <b>zero</b> scripts and the
+    /// database upgrade quietly does nothing.
+    /// </para>
+    /// <para>
+    /// These tests exist mainly to make the planned SDK-style project conversion (issue #510) safe, but
+    /// they guard any accidental rename just as well. If one fails, do not update the expected value
+    /// without first checking what loads that resource.
+    /// </para>
+    /// </remarks>
+    [TestClass]
+    public class EmbeddedResourceNameTests
+    {
+        /// <summary>
+        /// The prefix <see cref="DatabaseUpgrader"/> uses to discover SQL upgrade scripts. Kept as a
+        /// literal rather than referencing the private constant, so the test fails if the constant moves
+        /// away from the resources it is meant to match.
+        /// </summary>
+        const string SqlExtensionsPrefix = "App.ControlPanel.Engine.SqlExtentions";
+
+        static readonly string[] ExpectedControlPanelEngineResources =
+        {
+            "App.ControlPanel.Engine.Properties.Resources.resources",
+            "App.ControlPanel.Engine.SqlExtentions.Profiling-01-CommandExecute.sql",
+            "App.ControlPanel.Engine.SqlExtentions.Profiling-02-IndexOptimize.sql",
+            "App.ControlPanel.Engine.SqlExtentions.Profiling-03-CreateSchema.sql",
+        };
+
+        static readonly string[] ExpectedImporterEngineResources =
+        {
+            "WebJob.Office365ActivityImporter.Engine.ActivityAPI.Copilot.SQL.common_upsert_copilot_agents.sql",
+            "WebJob.Office365ActivityImporter.Engine.ActivityAPI.Copilot.SQL.insert_activity_from_staging_table.sql",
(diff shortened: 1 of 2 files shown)


sambetts:
## Review round 1 (grok-4.6) — one real hole, fixed in `35d10b51`

**should-fix (9/10): the prefix test did not actually pin `DatabaseUpgrader`'s constant.** It used its own literal copy of the prefix, so it only ever proved that three resources exist under that string — not that the running code still looks there.

The reviewer spelled out the bypass, and it is the exact failure the test was written to prevent:

1. Correct the `SqlExtentions` spelling in `SqlResourceNameStart` alone; leave the files where they are.
2. `DatabaseUpgrader_StillFindsItsSqlScriptsByPrefix` still finds 3 names under the **old** prefix. Green.
3. `ControlPanelEngine_EmbeddedResourceNames_AreUnchanged` still matches. Green.
4. At runtime `GetResourceNamesMatchingPathRoot` returns empty and the upgrade applies no scripts.

Now reads the private const by reflection, asserts it matches, and only then counts resources under it. It also fails with a clear message if the constant is renamed or moved.

Mutation-tested against precisely that scenario:

```
Failed DatabaseUpgrader_StillFindsItsSqlScriptsByPrefix
  Expected:<App.ControlPanel.Engine.SqlExtentions>
  Actual:<App.ControlPanel.Engine.SqlExtensions>
  DatabaseUpgrader's SQL resource prefix has changed.
```

The other four correctly stayed green. Restored: **5/5**.

## Coverage confirmed complete

The reviewer checked every other project for runtime resource loads and found **no uncovered path**. `Web`, `DataUtils`, `UsageReporting`, `CloudInstallEngine`, `WebJob.AppInsightsImporter.Engine` and `Tests.FakeDataGen` either have no embeds or load via `ResourceManager` (`Properties.Resources.*`), whose base name follows `RootNamespace.Properties.Resources` and is not the glob/path-rename failure mode.

One nuance worth recording: `insert_activity_from_staging_table.sql` is embedded and pinned here, but the importer actually reads it through `Resources.Insert_Activity_from_Staging_Table` (a resx file reference). Losing that embed would fail the exact-list test and **not** the importer — and would surface loudly as `ArgumentOutOfRangeException` rather than silently.

## EF snapshot test validated in detail

The reviewer confirmed the mechanism: the generated Designer does `new ResourceManager(typeof(TheMigration))`, so the blob is named after the **class** (`v106`), not the resx file (`201808231327312_v1-0-6`). The legacy `CreateCSharpManifestResourceName` + `DependentUpon` pairing is what produces that, and `UseDependentUponConvention` is exactly what the SDK conversion changes — which is the break this test exists to catch.

Also confirmed: every current `IMigrationMetadata` type is `public sealed partial` in `Common.Entities.Migrations` with a `.resx`; `Configuration` is correctly excluded; and a hypothetical code-only migration would not implement the interface, so it could not fail this test spuriously.

## Design choices confirmed as intended

- **Exact lists with `CollectionAssert.AreEqual`** rather than subset checks: a glob that *adds* a `.sql` must fail. Adding a legitimate new script should require updating the pin.
- **Hard-coded count of 3**: `>= 1` would not catch the zero-scripts case, which is the whole point.

Issue #528, opened 2026-09-12
EF Core on net10: current stable EF6 schema becomes the V1 baseline


sambetts:
## Decision

The `net10` branch **moves to EF Core** rather than carrying EF6 forward.

The current stable EF6 schema becomes **V1** — the baseline for the EF Core model. All existing EF6 migrations are assumed already applied; none of them are ported.

Future schema work:
- **Until .NET 10 goes to production** — EF6 migrations continue on `main`, and each must be mirrored as an EF Core migration on `net10` (see the dual-period risk below).
- **From the cutover** — EF6 migrations stop entirely; only EF Core migrations are added.

This supersedes the EF6-on-.NET-10 approach recorded in #511 and the epic #519. That approach was *proven to work* — EF 6.5.2 runs on .NET 10.0.11, migrations and all — but "it works" and "it is where we want to be" are different questions, and the decision is to stop carrying EF6 forward.

## What this makes easier

Collapsing **58 migrations** into a single baseline removes the entire class of problems that made the port awkward:

- No EF6 `.resx` gzipped model snapshots to preserve. That was the single biggest risk in #510 (SDK-style conversion) and a large part of why #527 exists.
- No `AutomaticMigrationsEnabled` / `AutomaticDataLossException` behaviour to reproduce.
- No `DbConfiguration` provider-name gymnastics (#511's three review rounds).
- `Microsoft.Data.SqlClient` is EF Core's native provider, so `Authentication=Active Directory Default` / `Managed Identity` work directly — the `AzureSqlTokenAuth` + `AzureSqlAccessTokenInterceptor` workaround can likely be **deleted** rather than ported.

## Measured surface to port

| Item | Count |
|---|---|
| `DbSet<>` declarations | 123 |
| `DbContext` types | 4 |
| `Database.SqlQuery<T>` call sites | 70 |
| `ExecuteSqlCommand` call sites | 41 |
| EF6 migrations | 58 → **collapsed to 1 baseline** |

Target: **EF Core 10.0.12** (`Microsoft.EntityFrameworkCore.SqlServer`, `.Design`).

## The baseline has to be exactly right, and that is the main risk

EF Core uses `__EFMigrationsHistory`; EF6 uses `__MigrationHistory`. They are different tables and can coexist, so an existing database is stamped by inserting the baseline row into `__EFMigrationsHistory` **without running it**.

[...]

PR #529, opened 2026-09-12
Add a schema snapshot and diff harness for database comparison

sambetts:
## Scope

Adds `SchemaSnapshot` (a test helper) and `SchemaSnapshotTests` (14 tests) to `Tests.UnitTests`. **No production code changes.** Two new files plus their `<Compile>` entries.

## Why

The EF Core baseline decision (#528) makes `net10` adopt EF Core with the current stable EF6 schema as **V1**. That whole approach rests on a single claim:

> a database created by the EF Core V1 baseline is **identical** to one created by applying every EF6 migration.

If that is not exactly true, the baseline is a lie and every later EF Core migration is a diff against the wrong starting point — producing changes that silently do the wrong thing on customer databases. There is currently no way to check that claim.

It is equally the guard for the **dual-migration period**. While `main` ships EF6 migrations and `net10` ships EF Core ones, every schema change has to be made twice. Comparing the two resulting schemas detects a missed change *regardless of whether anyone remembered to mirror it* — which is the only form of enforcement that works in practice.

## What it captures

| Category | Detail |
|---|---|
| Tables | schema-qualified names |
| Columns | type with **real** length/precision, nullability, identity, computed, collation, default |
| Indexes | uniqueness, primary key, type, **ordered** key columns, INCLUDE list, filter predicate |
| Foreign keys | referencing and referenced columns, delete/update actions |
| Check constraints | definition |

Output is a sorted set of canonical, self-describing strings rather than an object graph, because the useful product is a readable diff:

```
only in ef6: COLUMN dbo.users.user_name varchar(250) NOT NULL COLLATE SQL_Latin1_General_CP1_CI_AS
only in efcore: COLUMN dbo.users.user_name nvarchar(250) NOT NULL COLLATE SQL_Latin1_General_CP1_CI_AS
```

## What it deliberately ignores

Things that legitimately differ between two correctly-built databases: object ids, fragmentation, statistics, fill factor.

Two exclusions are judgement calls worth reviewing:

- **The EF history tables.** EF6 uses `__MigrationHistory`, EF Core uses `__EFMigrationsHistory` — by design. Reporting that would make every real comparison start with a false positive and get ignored. There is a test asserting it stays ignored.
- **Foreign key constraint *names*.** EF6 and EF Core auto-generate different names for the same relationship. A name difference is not a schema difference that affects the product; which columns reference which is what matters.

## Verification

```
Passed!  - Failed: 0, Passed: 14, Skipped: 0, Total: 14
```

Each test makes **exactly one** change to an otherwise identical pair of scratch databases and asserts the difference is both detected and named:

| Test | Why it is there |
|---|---|
| `IdenticalDatabases_ProduceNoDifferences` | no false positives |
| `NvarcharVersusVarchar_IsDetected` | the repo's hard Unicode rule — a baseline that got this wrong would be a **data-loss** bug |
| `ChangedIndexKeyOrder_IsDetected` | key order changes which plans an index can serve |
| `KeyColumnMovedToInclude_IsDetected` | measured in this repo at 5.5x wall-clock; a real behavioural change |
| `ChangedIndexFilter_IsDetected` | filtered indexes serve different queries |
| `MigrationHistoryTables_AreIgnored` | guards the exclusion above |
| plus added table/column, width, nullability, missing index, uniqueness, missing FK, changed default | |

## The tests found a real defect in the harness

First run failed all 14 with:

```
Cannot resolve collation conflict between "Latin1_General_CI_AS_KS_WS"
and "SQL_Latin1_General_CP1_CI_AS" in add operator
```

`sys` catalog columns are `sysname` and carry the *server* metadata collation, while string literals take the database default — so every concatenation threw. Fixed with explicit `COLLATE DATABASE_DEFAULT` on the catalog columns. Worth knowing for anyone else writing catalog queries in this codebase; several of the existing ad-hoc ones avoid it only by not concatenating.

## Secondary benefit

This consolidates a pattern the repo already reaches for constantly: roughly **57 one-off `sys.columns` / `sys.indexes` queries across 15 test files**, each checking a single fact. New migration tests can assert against a snapshot instead of hand-writing another catalog query.

## Reviewer notes

- Each test creates two scratch databases, so the class takes ~2 minutes. If that is too much for CI, the obvious lever is sharing one database pair across the read-only cases.
- Requires the `SPOInsightsEntities` connection string, same as the existing `ScratchDatabase`-based tests.
- `Tests.UnitTests.csproj` is still legacy format, so both files need explicit `<Compile>` entries.
- No database, network or customer data: every script is a synthetic `dbo.widgets` / `dbo.t`.

Relates to #528.

### DIFF
--- a/src/AnalyticsEngine/Tests.UnitTests/SchemaSnapshot.cs
+++ b/src/AnalyticsEngine/Tests.UnitTests/SchemaSnapshot.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Data.SqlClient;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 
 namespace Tests.UnitTests
 {
@@ -23,30 +24,31 @@ namespace Tests.UnitTests
     /// remembered to mirror it, which is the only form of enforcement that actually works.
     /// </para>
     /// <para>
-    /// The snapshot is a sorted set of canonical strings rather than an object graph, because the useful
-    /// output is a readable diff, not a tree. Every line is self-describing, so a failure message can be
-    /// pasted straight into a bug report.
+    /// The snapshot is a multiset of canonical strings rather than an object graph, because the useful
+    /// output is a readable diff. A multiset rather than a set because cardinality is meaningful: two
+    /// structurally identical foreign keys are not the same thing as one.
     /// </para>
     /// <para>
-    /// Deliberately captures only what the product depends on being right: tables, column definitions
-    /// (including nullability and collation), indexes with their key order and INCLUDE lists, foreign
-    /// keys, defaults, and check constraints. It ignores things that legitimately differ between two
-    /// correctly-built databases, such as object ids, index fragmentation, statistics and fill factor.
+    /// What it captures is driven by what this product's migrations actually do, not by completeness for
+    /// its own sake. Index options such as <c>IGNORE_DUP_KEY</c>, foreign-key trust state, and view bodies
+    /// are all in because migrations in this repository deliberately set them and depend on them. Things
+    /// that legitimately differ between two correctly-built databases - object ids, fragmentation,
+    /// statistics, fill factor, identity <c>last_value</c> - are deliberately out.
     /// </para>
     /// </remarks>
     internal sealed class SchemaSnapshot
     {
-        private readonly SortedSet<string> _lines;
+        private readonly List<string> _lines;
 
         public string Label { get; }
 
         private SchemaSnapshot(string label, IEnumerable<string> lines)
         {
             Label = label;
-            _lines = new SortedSet<string>(lines, StringComparer.Ordinal);
+            _lines = lines.ToList();
         }
 
-        public IReadOnlyCollection<string> Lines => _lines;
+        public IReadOnlyList<string> Lines => _lines;
 
         public static SchemaSnapshot Capture(string connectionString, string label)
         {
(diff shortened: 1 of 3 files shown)


sambetts:
## Review round 1 (gpt-5.6-sol) — three blockers, all real. Fixed in `6bdf9b0e`.

An excellent adversarial review. Each blocker was evidenced by a migration **in this repository** that deliberately sets the very thing the snapshot was ignoring — so these were not theoretical gaps.

### Blocker 1: `IGNORE_DUP_KEY` silently accepted

Two otherwise-identical unique indexes differing only in that option compared equal. `UniqueUrlsFullUrlIndex` sets it deliberately so one racing duplicate cannot abort an entire insert batch, and uses `ignore_dup_key = 1` as part of its target-state predicate. Also added `is_disabled` — a disabled index is not the same as an absent one.

### Blocker 2: views completely invisible

A database with a missing or stale view compared equal to one with the correct view. Views are part of the migration chain here: `v1-0-8` creates the reporting views and `RetireUnusedAuditYammerStreamTables` rewrites them.

Now captures views, procedures and functions. Bodies are whitespace-normalised **in C#** so reformatting is not reported while a real logic change is — there is a test for that, because a tool that cries wolf on every tidy-up stops being trusted. `uses_ansi_nulls` / `uses_quoted_identifier` are captured separately, since existing tests in this repo already treat quoted-identifier state as part of a view's required state.

### Blocker 3: untrusted constraints compared equal to trusted ones

`AddAuditEventsOperationFK` creates its foreign key `WITH NOCHECK` and then runs the expensive `WITH CHECK CHECK CONSTRAINT` specifically to reach `is_not_trusted = 0`. The snapshot would have certified the intermediate state as identical to the finished one — even though an untrusted constraint is unusable by the optimiser and may be hiding invalid rows. `is_disabled`, `is_not_trusted` and `is_not_for_replication` now captured for both foreign keys and check constraints.

### The three should-fix items, also done

| Finding | Fix |
|---|---|
| Cardinality lost to `SortedSet` — two structurally identical FKs compared equal to one | Comparison is now a **multiset**; duplicates report as a count difference. Constraint names stay excluded, since EF6 and EF Core auto-generate different ones |
| `IDENTITY(1,1)` and `IDENTITY(1000,5)` both rendered `IDENTITY`; computed columns lost their expressions | Both now carry their definitions. `last_value` still excluded — that is data state |
| History exclusion matched bare name in any schema | Qualified to `dbo`, with a test proving a history table in the wrong schema is still reported |

### Result

**26 tests, 0 failed** (was 14). Twelve new tests, each changing exactly one thing.

### Also addressed: the runtime concern

The review flagged ~2 minutes for the class. Reusing one pair of scratch databases across the class instead of creating two per test takes it from **3m33s to 11s** — database creation was the entire cost. Objects are dropped between tests in dependency order.

### One point I did not change

Excluding foreign-key constraint **names** from semantic equality. The review noted migrations that drop or guard constraints by exact name, which is true — but those are operations *on* a schema rather than properties *of* one, and including auto-generated names would make every EF6-versus-EF Core comparison fail on cosmetics, which is precisely the false-positive problem that gets a tool ignored. The multiset fix addresses the real risk the review identified, which was losing cardinality.

Issue #548, opened 2026-09-13
Copilot Adoption: freeze cohort membership when an intervention starts


sambetts:
## Summary

The enablement plan groups users by recommended action and lets the reader drill through to them - which is good, and is where any outcome measurement has to begin. But the grouping is **recomputed on every run**, so the membership changes as people's behaviour changes.

That makes later measurement impossible without selection bias: if you coach "everyone who is Trialling today" and measure "everyone who is Trialling in three months", you are measuring two different sets of people and the result means nothing.

**Whoever was actually treated has to be recorded at the moment of treatment.**

## Proposed shape

`dbo.copilot_adoption_cohort` (the cohort) and `dbo.copilot_adoption_cohort_member` (the frozen membership):

| Column | Notes |
| --- | --- |
| `cohort_id`, `name`, `action_code`, `created_utc`, `created_by` | |
| `baseline_period_end`, `baseline_options_hash` | so the cohort's starting state is reproducible |
| `user_id`, `baseline_band`, `baseline_score`, `baseline_active_days` | the member's state **at T0**, stored not recomputed |

Creating a cohort should be one action from the enablement-plan drill-through - "freeze this list as a cohort" - because that is the moment the user has decided to act on it.

## Acceptance criteria

- A cohort's membership never changes after creation.
- Each member carries its T0 state, so the starting point survives a later threshold change.
- A cohort can be exported (subject to the per-user access role, #538) so it can be handed to a department lead.
- Users can belong to more than one cohort over time, and a cohort can be marked closed.
- Migration is additive, idempotent and guarded, with a matching `.manual.sql`.

## Notes

- Depends on #542.
- Unblocks #549 (intervention record) and #550 (difference-in-differences).
- From the September 2026 Copilot Adoption review, plan item 3.1.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

Cohort membership is frozen at the moment of treatment.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

Issue #543, opened 2026-09-13
Copilot Adoption: show period-over-period movement on every headline figure


sambetts:
## Summary

There is no way to see whether adoption is improving. Every KPI is a point-in-time value with no comparison, and the only trend on the page counts raw active users rather than a rate (#545).

The expected executive question - *"adoption moved from 38% to 46% since last quarter, is that us or is that drift?"* - cannot be answered in the product at all.

## Proposed

Once #542 exists, add a delta to each headline figure: adoption rate, habit rate, reclaimable seats (by tier, #532), never-used, dormant, average and median engagement, unlicensed active users and recommended candidates.

Requirements that make it trustworthy rather than decorative:

- **Compare closed periods only.** Never the current partial period.
- **Honour the comparability gate.** If the two periods were scored under different `options_hash` values, either restate both from stored facts or refuse the delta and say why - never show a movement that is really a threshold change.
- **Show the denominator movement too.** An adoption rate can rise because people started using Copilot, or because seats were removed from people who weren't. Those are opposite conclusions and the rate alone cannot separate them. Report the seat-count change alongside.
- **Direction is not always good.** A fall in reclaimable seats is good; a fall in licensed users may or may not be. Do not colour a delta green purely because the number went down.

## Acceptance criteria

- Every headline KPI shows its prior-period value and the change, with the comparison period named.
- A threshold change between periods produces an explicit statement, not a silent delta.
- Seat-count movement is visible next to rate movement.
- Deltas appear in the workbook export as well as on screen.
- The comparison period is selectable (previous period / same period last quarter).

## Notes

- Depends on #542.
- From the September 2026 Copilot Adoption review, plan item 2.3.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

Every headline figure carries its period-over-period movement, gated by the `options_hash` comparability check so a threshold change can never be shown as a movement.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

Issue #542, opened 2026-09-13
Copilot Adoption: persist raw period facts so adoption history exists at all (foundation for progress tracking)


sambetts:
## Summary

The tool stores **no adoption history whatsoever**. Progress tracking today means exporting two Excel workbooks and subtracting them by hand - the workbook even ships a *"Comparing two snapshots"* method sheet explaining how.

This issue is the foundation for the whole milestone: a persisted, threshold-independent record of each closed period, from which bands, actions and rates are **recomputed on read**.

## Why not just recompute from the raw audit data?

Recomputing history on demand is attractive - no migration, no backfill problem, always current thresholds. It loses on three counts:

1. **Cost.** The single-period analysis already needs a 50,000-row cap, a 90-second query timeout, `MaxConcurrentSteps = 2`, a 10-minute cache and a lifecycle telemetry stream *because it has previously timed out* (#360, #441). Recomputing a dozen trailing windows interactively is a non-starter at the 200k-user baseline.
2. **It cannot reconstruct destroyed state.** Decisive. `UserLicenseProcessor` **deletes and rebuilds** `user_license_type_lookups` on every import (#277, #392), and Entra metadata is overwritten in place. Raw audit can never answer *"who held a seat then?"*, *"which department were they in then?"*, *"who was their manager then?"* - so a historical adoption **rate** has no denominator.
3. **Cohort progression needs identity.** *"Of the 300 never-used seats last quarter, how many are now active?"* requires last quarter's membership list. It cannot be recovered.

## Why not just store the score?

Because `CopilotAdoptionOptions` is explicitly tunable (wiki 2.7), and the moment someone tunes it the entire stored history becomes incomparable with the present - which is **worse than having no history**, because it still looks comparable.

## Proposed shape

**`dbo.copilot_adoption_user_period`** - one row per (period_end, period_days, user_id). **Raw facts only: no score, no band, no action.**

| Group | Columns |
| --- | --- |
| Period | `period_end`, `period_days`, `data_cutoff_utc` |
| Identity | `user_id` |
| **As-of state** (what recompute can never recover) | `seat_licence_type_ids`, `account_enabled`, `department_id`, `country_id`, `manager_id`, `seat_first_observed_utc` |
| Activity facts | `active_days`, `interactions`, `apps_used`, `agents_used`, `cowork_interactions`, `active_weeks`, `first_interaction_utc`, `last_interaction_utc`, `prior_interactions` |
[...]

Issue #277, opened 2026-08-19
Copilot: retain licence assignment history (seat history + assignment dates)


sambetts:
## Problem

Licence data is stored as **current state only**, with no history.

`UserLicenseProcessor` clears and rebuilds the lookups on every import:

```csharp
await db.Database.ExecuteSqlCommandAsync(
    $"DELETE FROM dbo.user_license_type_lookups WHERE user_id IN ({idList})");
```

So `user_license_type_lookups` always reflects the most recent import and nothing else. Once a licence is removed from a user, every trace that they ever had it disappears.

## What this makes impossible to answer

These are exactly the questions the Copilot reporting work leads to:

- **When was a user assigned their Copilot licence?** Adoption is meaningless without a start date — "used Copilot 3 times" reads very differently at week one than at month six.
- **How long did it take a user to become active after assignment?** Time-to-first-use is the core adoption metric and cannot be derived at all today.
- **How many seats were assigned over time?** Seat count is only ever "now", so growth, churn and reassignment are invisible.
- **Was a seat reclaimed from a dormant user and reassigned?** The licence-waste story needs this and it is unanswerable.
- **Did usage drop because people stopped using Copilot, or because seats were removed?** Two very different conclusions, currently indistinguishable.

## Proposed shape

A slowly-changing dimension alongside the existing current-state lookup, rather than replacing it:

- A history table recording user + licence SKU + first-seen + last-seen (and inferred removal), appended to per import rather than deleted and rebuilt.
- Graph exposes `assignedLicenses` but **not** an assignment date, so first-seen has to be inferred from the first import in which the licence appears. That means history only starts accruing from deployment, and the schema should make that limitation explicit rather than implying a true assignment date.
- Removal is likewise inferred from absence in a later import, so a failed or partial import must not be mistaken for a licence removal. This needs care.

## Scale note

Per the repository performance baseline, assume ~200,000 users. A history table that appends per user per import would grow unboundedly, so the write path must only record **changes**, not a row per user per cycle.

## Acceptance criteria

- Licence assignment and removal are retained after the fact rather than overwritten.
- Time-to-first-use can be calculated for a Copilot-licensed user.
- Seat counts over time can be reported.
- A failed or partial import cannot be misread as mass licence removal.
- The design is evaluated against the 200k-user baseline before implementation.

sambetts:
**Not picked up in this pass — this is a schema epic with a genuinely hard correctness problem at its centre.**

[...]

PR #579, opened 2026-09-16
Persist Copilot Adoption period facts

sambetts:
## Summary

Addresses #542.

This adds the Copilot Adoption period-history foundation: persisted, threshold-independent raw period facts plus read-time rescoring through the existing `CopilotAdoptionScoring` implementation. The interactive `AnalyseAsync` request path is unchanged; the new publish API is explicit/off-request and refuses current or missed historical periods so we do not backfill today's licence/department/manager state into the past.

## Architecture / code changes

- Added `dbo.copilot_adoption_user_period` for one raw fact row per `(period_end, period_days, user_id)`:
  - period/data cutoff
  - user id and as-observed seat licence type ids
  - account/department/country/manager ids
  - honest `seat_first_observed_utc` based only on `users.created_utc` (NULL when unknown; no #277 seat-history inference)
  - audit/report activity facts and `coverage_status`
- Added `dbo.copilot_adoption_period_run` with `options_hash`, coverage/provenance and publication counts.
- Added `CopilotAdoptionService.PublishClosedPeriodAsync(...)`:
  - publishes only the latest closed period (`yesterday` UTC) to avoid manufacturing historical org state
  - blocks overlapping periods of the same length
  - deletes/re-inserts the same `(period_end, period_days)` inside a transaction for idempotent re-publish
- Added `ReadPublishedPeriodAsync(...)` to read stored facts and recompute bands/actions/rates using the existing scorer under the current options.
- Added `ComparePublishedPeriodsAsync(...)` comparability gate; differing `options_hash` values are flagged instead of treated as directly comparable.
- Added period model DTOs and SQL integration coverage for live-vs-stored scoring, idempotent re-publish and option-hash mismatch.
- Updated `Create DB.sql`, `Entities.csproj`, manual migration assets, and `UrlFullUrlMigrationPipelineTests.LatestId`.

## Schema effects / migration classification

Migration: `202609170900001_CopilotAdoptionPeriodFacts`.

Classification: **purely additive**.

- Adds two new tables, table-local constraints, a unique run index and the requested clustered period/user fact index.
- No existing table is rewritten and no existing query/index is being tuned.
- The clustered fact index is additive with the new empty table and **unmeasured** because there is no before-plan to compare; it is not a performance-motivated change to existing schema.
- The EF model is unchanged; `.resx` is copied verbatim from `202609151440027_CopilotPromptSafetyFields` and the manual script stamps by copying that predecessor row.
- Manual SQL script included: `202609170900001_CopilotAdoptionPeriodFacts.manual.sql`. Pre-stamp guard checks schema only, never row data. No mop-up pass is needed because the migration does not backfill.

Upgrade/runtime: creating empty tables and indexes is expected to be near-instant; no maintenance window is required beyond normal schema-upgrade caution.

## Tests / validation

Attempted:

- `MSBuild.exe src\AnalyticsEngine\Common\Entities\Entities.csproj -t:Restore,Build -p:Configuration=Release -p:Platform=anycpu -v:m`
  - VS18 initially cannot resolve `Microsoft.NET.Sdk` for `DataUtils` unless `MSBuildEnableWorkloadResolver=false` and `MSBuildSDKsPath=C:\Program Files\dotnet\sdk\10.0.400\Sdks` are supplied.
- With that environment workaround, restore succeeds but build is blocked before this change can be validated by an existing local assembly conflict:
  - `CS1705: Assembly 'Microsoft.Data.SqlClient' ... uses 'System.Data.Common, Version=4.2.0.0' which has a higher version than referenced assembly 'System.Data.Common, Version=4.1.1.0'`.
- `Tests.UnitTests` restore succeeds with the same environment workaround.
- `Tests.UnitTests` build is blocked locally because the installed BuildTools layout lacks `Microsoft.WebApplication.targets` for both VS18 and VS2022 BuildTools.
- `git diff --check` passed.

This PR is draft until CI or a workstation with the expected VS web build targets confirms the targeted Copilot Adoption tests.

## Risks / reviewer hotspots

- `PublishClosedPeriodAsync` intentionally refuses older missed periods. That is conservative, but it is what prevents the #542 anti-requirement: current licence/department/manager state must not be written into historical periods.
- `seat_first_observed_utc` is an honest degraded signal. It uses `users.created_utc` where available and otherwise remains NULL; true Copilot seat-assignment history remains deferred to #277.
- The read path joins stored ids back to current dimension labels for display, but the persisted facts retain the ids captured at publish time.
- The option hash gates direct period comparisons; future UI work should surface the message and, where needed, restate both periods from raw facts under one option set.
- Other parallel branches also add migrations, so a conflict on `LatestId` in `Tests.UnitTests/UrlFullUrlMigrationPipelineTests.cs` is expected at merge.

## Deferred work

- UI/API endpoints for period-over-period deltas, cohort movement, trends and target tracking remain for #543-#548.
- A real scheduler/web-job trigger that calls `PublishClosedPeriodAsync` is still needed; this PR adds the publish/read foundation and keeps it off the interactive request path.
- True seat assignment history and historical department/manager state remain out of scope until #277 lands.

## Wiki updates required

Do not edit the shared wiki checkout in this PR. Proposed wiki changes for `Copilot Adoption Tool.md`:

1. Add a new section, "Persisted period facts and progress tracking", explaining that the tool now persists raw period facts for closed periods and recomputes all bands/actions/rates on read.
2. State that stored facts are threshold-independent: scores, bands, actions and rates are never stored because `CopilotAdoptionOptions` is tunable.
3. Document the comparability rule: periods with different `options_hash` values must be restated under one option set before being compared.
4. Document coverage semantics: `coverage_status = unverifiable` means show a gap/unknown, not zero adoption.
5. Document the limitation until #277: `seat_first_observed_utc` is populated only from genuinely available lifecycle/first-observation data and may be NULL; the product does not reconstruct historical licence, department or manager assignments.
6. Add an operator note that publishing is scheduled/off-request and only closed periods are eligible; missed historical periods are deliberately not backfilled from current state.


sambetts:
Fixed the PR #579 publish/read-path findings in `2cf692dc`.

| Finding | Status | Fix |
|---|---|---|
| BLOCKER 1 | Fixed | `HasCopilotAuditDataSql` is now bounded by `@from <= time_stamp < @toExclusive`; publish probes pending Copilot backfill first and downgrades stored coverage to `unverifiable`/report-only rather than writing verified zero audit adoption. Added no-in-window-audit and pending-backfill regression tests. |
| BLOCKER 2 | Fixed | publish execution now uses its own non-interactive command timeout (`CommandTimeout = 0`) instead of the 90-second request-path timeout. |
| MAJOR 3 | Fixed | stored periods are rescored with a cloned options object whose `WindowDays` is the persisted `period_days`; the summary period end is no longer mixed with the exclusive cutoff. Added a D28-read-under-D90-options regression test. |
| MAJOR 4 | Fixed | `PublishedPeriodFactsSql` now re-joins the current reclaim exclusion inputs with the same active/expired semantics as the live read path. Added exclusion-seeded stored-vs-live coverage. |
| MAJOR 5 | Fixed | stored `seat_licence_type_ids` are still returned separately, while `SeatLicences` now resolves those ids to `dbo.license_types.name` for display/CSV parity. |
| MAJOR 6 | Fixed | extended the existing additive migration/manual script/Create DB definition with `account_created_utc`; publish leaves `seat_first_observed_utc` NULL until real seat history exists and stores Entra account creation separately. Manual Up SQL still matches the migration constant. |
| MAJOR 7 | Fixed | `PublishClosedPeriodAsync` now returns `Published`, `AlreadyPublished`, or `NotDue` for expected scheduler states; exceptions are reserved for real faults. Added exact-retry/current/older/overlap status tests. |
| MINOR 8 | Fixed | persisted `report_period_days` on the run row and hydrate it into stored-read data sources so report-window mismatch logic still works. |
| MINOR 9 | Fixed | stored-read `Summary.ToUtc` now uses the inclusive `period_end`, while `DataCutoffUtc` remains on the run row as the exclusive cutoff. |
| MINOR 10 | Fixed | `OptionsHash` now covers only fact-shaping options (`WindowDays`, `HistoryDays`, `UsageReportLagDays`) instead of every scoring/display threshold. |
| MINOR 11 | Not fixed | Left as deferred because the current schema does not persist job title / office / company ids; fixing it honestly would require more historical fact columns beyond the requested blocker/major scope. |

Validation:

- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -t:Restore -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `vstest.console.exe src\AnalyticsEngine\Tests.UnitTests\bin\Release\Tests.UnitTests.dll /TestCaseFilter:"FullyQualifiedName~CopilotAdoption|FullyQualifiedName~Period"` — 259/259 passed
- `git diff --check`
- Verified `202609170900001_CopilotAdoptionPeriodFacts.manual.sql` embeds `CopilotAdoptionPeriodFacts.Up_Sql` verbatim.

Issue #541, opened 2026-09-13
Copilot: migrate to the /copilot Graph usage API and import the v2 fields


sambetts:
## Summary

The Copilot usage-report import targets the Graph **beta** `/reports/` endpoint. Microsoft has superseded it, verbatim from the reference page:

> "**Going forward, use the Microsoft 365 Copilot usage APIs under the `/copilot` URL path segment.**"

The replacement is **GA on v1.0**:

```
GET https://graph.microsoft.com/v1.0/copilot/reports/getMicrosoft365CopilotUsageUserDetail(period={period}, version={version})
```

Three methods exist: `getMicrosoft365CopilotUsageUserDetail`, `getMicrosoft365CopilotUserCountSummary`, `getMicrosoft365CopilotUserCountTrend`.

## Why it is worth doing now

`version=v2` carries fields we do not currently have, and which the adoption tool would use directly:

- **Prompts submitted** for all apps, and split for Copilot Chat (work) and Copilot Chat (web)
- **Active usage days for all apps**
- Microsoft 365 Copilot, Edge and **Copilot Agent** last-activity dates

Today `ReportAppsUsed` has to be *derived* from per-app last-activity dates (`CopilotAdoptionSql.ReportAppsUsedExpression`, `CopilotAdoptionSql.cs:400+`), with a documented workaround for Microsoft having split the single "chat" date into work and web between report versions. The v2 fields reduce that inference.

It also matters for the fallback path: when the audit import is unavailable or lagging, Microsoft's report is the **only** signal, and it decides whose seat is flagged for reclaim (`CopilotAdoptionScoring.cs:104-138`). Better fields there directly improve reclaim safety.

## Known limit to design around

> *"This API only returns usage data for users who have a Microsoft 365 Copilot license. **Unlicensed Copilot Chat usage data isn't available through Microsoft Graph reports APIs.**"*

So this **supplements** the audit-log import; it cannot replace it. Unlicensed Copilot Chat demand - the basis of the entire licence-opportunity feature - remains reachable only via the audit log / Management Activity API. That asymmetry is worth stating in the same change (see #535).

## Acceptance criteria

- The importer targets the v1.0 `/copilot` path with a documented fallback while tenants are mid-rollout.
- v2 fields are imported and available to the adoption scorer; `ReportAppsUsed` prefers the reported value over the derived one when present.
[...]

PR #580, opened 2026-09-16
Migrate Copilot usage reports to Graph v1

sambetts:
## Summary

Addresses #541.

Migrates the Microsoft 365 Copilot usage-report importer from the legacy beta reports endpoint to the GA Microsoft Graph v1.0 `/copilot/reports` path and imports the v2 per-user usage fields needed by Copilot Adoption scoring.

## Code changes

- Adds GA CSV-stream ingestion for the three Copilot reports:
  - `getMicrosoft365CopilotUsageUserDetail`
  - `getMicrosoft365CopilotUserCountSummary`
  - `getMicrosoft365CopilotUserCountTrend`
- Keeps a documented mid-rollout fallback chain:
  1. v1.0 `/copilot/reports`, requested version (`v2` by default)
  2. beta `/copilot/reports` JSON `version='v2'`, preserving the pre-#541 v2 endpoint for tenants that have not rolled out the GA CSV stream yet
  3. v1.0 `/copilot/reports`, `v1` with `D28` mapped to `D30`
  4. legacy beta `/reports` JSON path as the final v1 fallback
- Converts GA CSV streams back into the existing JSON-shaped parser boundary so the existing narrow/tall aggregate model and per-user persistence rules remain in place.
- Records successful load provenance from the actual fallback used (`report_version`, `report_period`) in the import log.
- Adds per-user snapshot column `report_version` (`nvarchar(10) NULL`) so per-user snapshot facts trace back to the report schema that produced them
- Imports v2 prompt counters, active-usage days, Microsoft 365 Copilot / Edge / Copilot Agent / split Chat last-activity dates.
- Keeps `CopilotAdoptionSql.ReportAppsUsedExpression()` derived from documented per-app last-activity dates; chat, chat work, and chat web still collapse to one comparable chat surface.
- Preserves the current safety behaviour for v1/fallback responses: if v2 values are absent, existing v2 counters are not blanked.

## Schema / migration effects

Migration: `202609170910001_GraphCopilotUsageApiV2`.

Classification: **additive**. It adds one nullable column to `dbo.copilot_usage_user_activity_log`; it does not rewrite existing rows, remove schema, or add a performance-motivated index. Upgrade cost is metadata-only and should be effectively instant regardless of tenant size.

A matching manual script is included and uses schema-only pre-stamp guards. It stamps `__MigrationHistory` with this migration's EF model snapshot after verifying the `report_version` column exists.

`UrlFullUrlMigrationPipelineTests.LatestId` was updated to the new migration id. A merge conflict in that file is expected if parallel branches add later migrations.

## Tests

Added coverage for:

- GA v1.0 CSV v2 user-detail parsing, including prompt counts, active days, Edge, Microsoft 365 Copilot, Copilot Chat work/web and Copilot Agent last activity.
- GA-v2 to beta-v2 fallback, then v1 fallback path and provenance recording.
- `ReportAppsUsed` deriving breadth from documented per-app last-activity dates after dropping the speculative Graph app-breadth field.

Validation attempted:

- `git diff --check` passes.
- `MSBuild src\AnalyticsEngine\Tests.UnitTests -t:Restore ...` passes.
- `MSBuild src\AnalyticsEngine\Common\Entities\Entities.csproj -t:Restore,Build ...` is blocked in this workstation by a pre-existing assembly resolution error: `CS1705` for `Microsoft.Data.SqlClient` requiring `System.Data.Common, Version=4.2.0.0` while MSBuild chooses `System.Data.Common, Version=4.1.1.0` from the transitive package asset.
- `MSBuild src\AnalyticsEngine\Tests.UnitTests ...` is separately blocked because this BuildTools installation is missing `Microsoft.WebApplication.targets` under `v18.0\WebApplications`.

Because local validation is blocked by build-tooling/environment issues rather than a test failure in this change, this PR is opened as a draft.

## Risks / reviewer hotspots

- Review `CopilotReportCsvParser` header mappings against the current Microsoft Learn/admin-center labels. Microsoft documents the v2 field display names but the GA v1.0 response is a CSV stream, so the importer maps documented CSV headers into the existing parser property names.
- Review the fallback semantics: only 400/404/not-found advance to fallback; other Graph errors still fail so permission/auth problems do not look like empty reports. The beta `/copilot` v2 JSON endpoint is tried before the v1 downgrade.
- Review `report_version` handling for fallback snapshots. New v1 rows are stamped `v1`; existing v2 facts are not downgraded when a later v1 fallback lacks v2 values.

## Known product limit

The Microsoft Graph Copilot usage reports API only returns users who have a Microsoft 365 Copilot licence. Unlicensed Copilot Chat usage is not available through Graph reports APIs, so this supplements the audit-log / Management Activity API import and cannot replace it. Unlicensed demand for licence-opportunity analysis remains audit-log-only.

## Wiki updates required

Do not edit the shared wiki checkout from this PR. Proposed wiki changes:

- Update the Copilot Adoption / importer documentation to state that Microsoft 365 Copilot usage reports now use `https://graph.microsoft.com/v1.0/copilot/reports/...` with a v1/legacy fallback during rollout.
- Document that Graph report data covers licensed Copilot users only; unlicensed Copilot Chat demand remains available only from the audit-log import.
- Add the new imported v2 fields and explain that Report Apps Used is derived from documented per-app last-activity dates, with chat work/web collapsed to one chat surface for v1/v2 comparability.
- Add manual migration script `202609170910001_GraphCopilotUsageApiV2.manual.sql` to release-operator upgrade notes when this reaches stable.


sambetts:
Fixed the review findings in cdc4cb79.

| Finding | Status | Fix |
|---|---|---|
| BLOCKER 1 | Fixed | Restored the pre-#541 `beta/copilot/reports/...version='v2'?$format=application/json` fallback before any v1 downgrade. The chain is now GA v1.0 CSV v2 -> beta `/copilot` JSON v2 -> GA v1.0 CSV v1 -> legacy beta `/reports` JSON v1, with provenance set from the successful attempt. |
| MAJOR 2 | Fixed | Added the documented `Edge Last Activity Date` CSV header mapping and changed the v2 CSV fixture to Microsoft Learn's documented labels. |
| MAJOR 3 | Fixed | Dropped the speculative `apps_used` Graph field entirely: parser aliases, row/entity property, loader write, SQL `COALESCE`, scratch schemas, migration column, `.resx` Target, and manual-script model blob were removed/updated. `ReportAppsUsed` remains derived from documented per-app last-activity dates. |
| MAJOR 4 | Fixed | Added a `ResponseHeadersRead` stream path for GA CSV downloads and changed `CopilotReportCsvParser` to read records incrementally, avoiding the raw CSV string + full records list + JObject list triple-buffering. |
| MINOR 5 | Fixed | Aggregate enabled/active user headers are now derived safely from the raw/normalized prefix instead of calling `Substring(0, -1)` on renamed headers. |

Validation:

- Microsoft Learn checked for the documented v1.0/beta `/copilot/reports` pivots and the v1/v2 user-detail column labels, including `Edge Last Activity Date`.
- `git diff --check`
- `MSBuild src\AnalyticsEngine\Tests.UnitTests -t:Restore -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `MSBuild src\AnalyticsEngine\Tests.UnitTests -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `vstest ... /TestCaseFilter:"FullyQualifiedName~CopilotUsageReportImport|FullyQualifiedName~CopilotAdoptionSqlIntegration|FullyQualifiedName~UrlFullUrlMigration"` passed 48/48.

Migration consistency checks: the `202609170910001_GraphCopilotUsageApiV2.resx` Target no longer contains `apps_used`/`AppsUsed`, and the manual script's stamped model blob byte-matches that Target.

sambetts:
## Release-assembly note: migration snapshot ordering across the four parallel migrations

Flagging a cross-PR hazard found while integrating this release. **No action is needed on this PR in isolation** — it is correct as it stands — but it must be checked once the migration-bearing PRs are merged.

### The situation

This release currently carries four new migrations, on top of `202609161200001_IndexTeamsExplorerQueries` which is already on `dev`:

| Run order (migration id) | PR | Changes the EF model? | How its `.manual.sql` stamps |
|---|---|---|---|
| `202609170900001_CopilotAdoptionPeriodFacts` | #579 | **No** — new tables are not `DbSet`s | copies predecessor's `Model` |
| `202609170910001_GraphCopilotUsageApiV2` | #580 | **Yes** — adds `report_version` | embeds its own freshly scaffolded blob |
| `202609170920001_CopilotSubscribedSkuCapacity` | #583 | **No** | copies predecessor's `Model` |
| `202609170940001_CoworkUsageReportTables` | #588 | **No** | copies predecessor's `Model` |

Verified by decoding each `.resx` `Target` and comparing against the `dev` baseline: only #580's differs (baseline 46,122 bytes).

### The hazard

Each branch's snapshot is correct *relative to its own branch*, because each was cut from a `dev` that did not contain the other three. But the migrations are applied in **id order**, and #580 — the only one that changes the model — sits **second**. The two raw-SQL migrations that run after it (`...0920001`, `...0940001`) copy a predecessor row whose `Model` predates `report_version`.

The consequence is not in the schema — every table and column is created correctly — it is in what `__MigrationHistory` *claims* the model is. If the highest-id migration ends up stamped with a pre-`report_version` model while the built entity model contains it, EF sees the schema as behind the build. Concretely that surfaces as:

- `HealthServiceTests.Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` failing in CI, and
- at runtime, `Migrations/Configuration.cs` having `AutomaticMigrationsEnabled = true`, so EF would auto-generate an on-the-fly migration rather than treating the schema as current.

This only bites when more than one of these four merges, which is why no individual PR shows it.

### What to check at merge time

After the migration-bearing PRs are merged to `dev`:

1. Confirm the **highest-id** migration's snapshot matches the final merged entity model. If #580 has merged, that snapshot must include `report_version`.
2. Treat `Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` as the gate — it is the check that actually proves it, and it runs in CI. (It cannot be run on a developer machine: `App.Release.config` ships the unsubstituted placeholder `__TenantGUID__`, so that whole test class fails in `Guid.Parse` locally. ~161 such failures appear on every branch, including untouched ones.)
3. If it fails, re-scaffold the highest-id migration's `.resx` against the merged model and update its `.manual.sql` stamp to match, rather than relaxing the test.

### Why the manual scripts themselves are still safe

Each `.manual.sql` names a predecessor that is already on `dev`, so the severity-16 prerequisite guard cannot spuriously fire regardless of the order these four land in. The stamp guards were separately verified to be **schema-only** (`OBJECT_ID` / `COL_LENGTH` / `sys.indexes`), with the already-applied path falling through to the stamp rather than returning early — so re-running any of them is a no-op, and none of them can refuse to stamp because of data state.

Issue #552, opened 2026-09-13
Copilot Adoption: split the Overview into an Executive view and an Analyst view


sambetts:
## Summary

The Overview carries roughly 16-18 distinct visualisations. That is an excellent **analyst workbench** and a poor **executive briefing**, and the stated primary audience is now C-level.

The wiki already documents the fix that was applied once before - an earlier draft *"ran seventeen cards in the order they happened to be built, which buried 'adoption by department' - the *who do I talk to* card - near the bottom"*. The four numbered sections solved the ordering problem but not the volume problem.

## Proposed

**Executive view - three acts, one screen:**

1. **Where we stand** - seats, adoption rate **with its period delta** (#543), habit rate, **reclaim split by confidence** (#532), idle spend (#553), reassignment opportunity.
2. **Where it is working and failing** - department league table ranked by habit rate showing seat surplus/deficit, plus the funnel.
3. **What we are doing about it** - the enablement plan with owners (#549), and last period's cohort outcomes (#544).

**Analyst view** - everything else, unchanged: intensity scatter, score-profile radar, concentration cohorts, resource types, treemaps, country breakdown, and the SQL popovers.

## On the InfoTips and the SQL popover

One reviewing model argued these are engineering reassurance aimed at the wrong audience and should go. **Disagree in part.** Wiki principle 2.1 (*"every assertion must explain itself, on the spot"*) is a genuine differentiator and the reason these numbers survive a licence negotiation - the tips should stay.

The **placement** is what is wrong for an executive: keep the one-line definition inline, and move the `</> SQL` popover to the Analyst view. A `</> SQL` button beside a board-pack KPI reads as unfinished software to a C-level audience, while being exactly right for the admin verifying a figure.

## Acceptance criteria

- The Executive view answers "where do we stand / where is it failing / what are we doing" without scrolling past unrelated diagnostics.
- No figure exists only in the Executive view - every headline drills through to its detail, per wiki 2.9.
- The SQL popover remains available in the Analyst view for every query-backed figure (wiki 2.5 is not weakened).
- Colour still carries meaning a second way in both views (wiki 2.8).
- The empty state (wiki 2.10) is designed for the Executive view too.
- Wiki section 5 rewritten in the same change.

## Notes

[...]

Issue #545, opened 2026-09-13
Copilot Adoption: the weekly trend applies today's seat list to six months of history


sambetts:
## Summary

The weekly trend is the tool's only in-product answer to *"is our enablement programme working?"*, and it can show false growth, false decline, or an artificial collapse - for three independent reasons.

## 1. Retrospective licence attribution

`CopilotAdoptionSql.WeeklyAdoptionTrendSql` (`CopilotAdoptionSql.cs:1071`) resolves the licensed population from the **current** `user_license_type_lookups` and applies it across the whole six-month range (`TrendMonths = 6`, `CopilotAdoptionService.cs:68`).

So:

- a user licensed **today** is counted as licensed throughout their entire prior history;
- a user whose licence was **removed** disappears from history retrospectively;
- the series is a raw count with **no historical seat denominator**, so it cannot express a rate at all.

The effect is systematic, not random: a tenant that has been steadily adding seats will see a flattering historical curve, because today's larger population is projected backwards.

## 2. The current, incomplete week is included

The spine runs to the latest Monday, so the most recent point is a partial week and always reads as a drop. Anyone opening the page mid-week sees a decline that is purely an artefact of when they looked.

## 3. Missing weeks are filled with zero

`FillWeeks` (`CopilotAdoptionService.cs:1814`) fills gaps with `0`, deliberately:

> *"Zero (rather than a gap) is right here: these series count audit events, and no events genuinely does mean nobody used Copilot that week."*

That reasoning is **correct for a genuine zero and wrong for an import outage**, and the two are indistinguishable here. `AdoptionTimePoint.Value` is already `double?` precisely so a gap can be drawn - the capability exists and is unused for this case.

## Suggested direction

Once #542 publishes closed periods with as-of seat state:

- compute the trend against the **as-of licensed population** for each period, making a genuine adoption *rate* possible;
- **exclude the incomplete period** entirely;
- emit `null` for periods whose import coverage cannot be verified, and draw them as gaps with a legend note.

An interim improvement available before #542: exclude the current partial week, and state on the chart that the licensed population is as-of-today rather than as-of-then.

## Acceptance criteria

- No point in the series uses a licence population from a different date than the activity it plots.
- The current partial period is never plotted.
- Unverifiable coverage renders as a gap, and the chart says what a gap means.
- The chart can show a rate, not only a count.

## Notes

- Depends on #542 for the as-of denominator.
[...]

PR #584, opened 2026-09-16
Fix Copilot adoption weekly trend gaps

sambetts:
Addresses #545

## Summary
- Excludes the current partial week from the Copilot Adoption weekly trend by passing an exclusive `@trendTo` into `WeeklyAdoptionTrendSql` and building the chart spine only from completed Mondays.
- Adds a weekly Audit.General coverage probe and threads it into `FillWeeks`: a missing trend row is now `0` only when that completed week has import evidence; otherwise it remains `null` so the existing nullable time-point model draws a gap.
- Updates the portal and workbook caveats to state the interim limitation clearly: licensed/unlicensed membership still uses today's `user_license_type_lookups`, not as-of-then seat state.
- Prevents the stacked composition chart from turning nullable coverage gaps into a false collapse; the line charts show those weeks as gaps with an in-chart legend note.

## Deferred to #542
This branch deliberately does **not** add period-fact tables or an as-of-then denominator. Today's licence lookup is still used for historical activity attribution, but the chart now says that explicitly. #542 can replace the coverage probe / current-seat lookup with closed-period facts and a true historical rate without unpicking portal rendering, because the API already carries nullable `AdoptionTimePoint.Value` gaps.

## Tests and validation
- Added C# regression coverage for:
  - mid-week current partial period exclusion;
  - Audit.General coverage probing instead of `copilot_chats` coverage;
  - unverifiable missing weeks returning `null`;
  - verified no-activity weeks returning `0`;
  - explicit SQL zero rows remaining `0`.
- Added portal chart test proving a nullable point starts a new SVG line segment and renders the gap legend note.
- `git diff --check` ✅
- Portal `node node_modules\\vitest\\dist\\cli.js run --reporter=dot` ✅ (all portal tests passed)
- Portal `node node_modules\\typescript\\bin\\tsc --noEmit` ✅
- Requested C# validation is blocked on this worker image:
  - `MSBuild.exe src\\AnalyticsEngine\\Common\\Entities\\Entities.csproj -t:Restore,Build ...` fails before this change is compiled because `Common\\DataUtils\\DataUtils.csproj` cannot resolve `Microsoft.NET.Sdk` from the VS18 BuildTools MSBuild SDK folder.
  - `MSBuild.exe src\\AnalyticsEngine\\Tests.UnitTests ...` restores successfully, then fails because `Microsoft.WebApplication.targets` is not installed under the VS18 BuildTools path.
  - With a local SDK workaround, `Entities.csproj` reaches C# compilation and fails on a pre-existing assembly-resolution issue (`Microsoft.Data.SqlClient` requiring `System.Data.Common` 4.2.0.0 while 4.1.1.0 is selected); no new compile errors are reported before that blocker.

## Both-direction regression evidence
- The old trend spine was `WeekSpine(trendStart, MondayOf(DateTime.UtcNow.Date))`, so the new `WeeklyTrend_FiltersOutTheCurrentPartialWeek` assertion would have received the current Monday and failed.
- The old `FillWeeks` returned `0` for every missing row, so `FillWeeks_UsesNullForUnverifiableCoverage` would fail against the old implementation.
- `FillWeeks_KeepsZeroForVerifiedNoActivityWeeks` and the explicit-zero test guard the opposite direction so genuine zeros are not converted into gaps.

## Risks / reviewer hotspots
- Coverage is an interim Audit.General evidence check. A quiet tenant week with no General-feed evidence is intentionally shown as unknown rather than guessed as zero.
- The coverage query scans the bounded trend window against `audit_events.time_stamp` and checks `event_meta_general` existence; reviewers should check the plan on large tenants if #542 does not replace this soon.
- Portal overlap with #552: changes are intentionally confined to the existing trend cards and generic `TimeSeriesChart` gap legend prop.

## Wiki updates required
- Update the Copilot Adoption methodology page to say weekly trend points exclude the current partial week.
- Document that gaps mean Audit.General import coverage could not be verified for that completed week, not zero activity.
- Document the interim licence caveat: historical trend membership is as-of-today until #542 ships closed periods with as-of seat state and a true rate.


sambetts:
Fixed the four review findings in this branch.

- Volume trend workbook chart now switches from stacked area to line when any exported weekly volume point is blank, matching the portal's blank-cell safety rule.
- Weekly coverage probing is now driven from a completed-week spine with `CROSS APPLY TOP (1)`, so it performs one seek-and-stop existence check per week instead of aggregating every Audit.General row in the six-month window.
- The trend spine now clips leading weeks before the tenant's first audit evidence while preserving interior unverifiable weeks as gaps, so new installs and short-retention databases do not hide the composition chart for months.
- Customer-facing portal/workbook strings now use plain English for the closed-period seat-snapshot limitation and no longer expose the internal issue number.

Validation:
- Enterprise MSBuild Release build for `src\AnalyticsEngine\Tests.UnitTests`: passed.
- `vstest.console.exe /TestCaseFilter:"FullyQualifiedName~CopilotAdoption"`: 239 passed.
- `npm ci --loglevel verbose --timing --no-audit --no-fund`: passed after the normal `npm ci` run was silent under load.
- Full `npm run test`: 208 passed, 1 failed in unrelated `src/pages/LicenceActivityPage.test.tsx` at the department option assertion.
- Targeted Copilot Adoption portal tests (`src\components\copilotAdoption\*.test.tsx` resolved explicitly): 37 passed.
- `npx tsc --noEmit`: passed.

Issue #553, opened 2026-09-13
Copilot Adoption: put a currency figure on idle licence spend (and only on idle licence spend)


sambetts:
## Summary

`ReclaimableSeats` is a raw count. The number a budget owner actually needs is what those seats **cost**, and the tool has no cost dimension at all.

## Scope - deliberately narrow

**Do:**

- One input: **annual (or monthly) seat cost + currency**, per Copilot SKU.
- **Idle licence spend** = *eligible* reclaimable seats x cost, carrying the same eligibility caveats and confidence tiers as the count (#532).
- Report **spend exposure**, **reassignable** and **reducible-at-renewal** as three separate figures. They are different decisions: a seat you can give to someone else is not a saving, and conflating them is how a tool loses a CFO's trust.

**Do not:** ROI, "hours saved", or a renewal forecast. Not derivable from interaction logs. All three models reviewing this area independently advised against it, and a fabricated ROI would discredit the true numbers beside it. See #551 for the honest treatment of value.

## Reuse the existing convention

This product already ships currency-aware cost reporting in the Agent costs area - `AzureCostByCurrency` carries `currency` and `cost` and is documented as *"Never added across currencies."* Reuse that model and that rule rather than inventing a second one.

## Sequencing matters

**This must come after #532 and #537.** Multiplying a wrong count by a currency symbol makes a wrong, named, adverse claim look authoritative. Idle spend is only as trustworthy as the eligibility filtering underneath it.

## Acceptance criteria

- Cost is optional; with no cost configured the tool shows counts exactly as it does today, with no empty currency fields.
- Figures never sum across currencies.
- Idle spend is broken down by the reclaim confidence tiers, so the *Certain* figure can be quoted on its own.
- Reassignable and reducible are separate figures with distinct definitions stated at the number.
- The cost input and its effective date are recorded, so a historical figure can be traced to the price used.
- Appears in the workbook.

## Notes

- Market context: chargeback/showback is **table stakes** here, not a differentiator - CoreView, Rencore and AvePoint Cense all ship it, and Microsoft now ships per-user/group cost visibility for Cowork credits. We ship none. This is catching up, not gold-plating.
- From the September 2026 Copilot Adoption review, plan item 4.3.

sambetts:
Not delivered by **Stable build 1833**, and deliberately left open.

[...]

Issue #549, opened 2026-09-13
Copilot Adoption: record the intervention, so the "enablement plan" becomes an actual plan


sambetts:
## Summary

The enablement plan tells you *how many people need each action*. It does not record that anyone did anything about it.

There is no owner, no date, no intervention type, no status and no follow-up - so the plan is a **categorised queue**, not a plan, and principle B of this tool ("groups map to concrete actions, plans and outcomes") is only half met.

## Proposed shape

An intervention record attached to a frozen cohort (#548):

| Field | Notes |
| --- | --- |
| `cohort_id` | who was treated |
| `owner` | the person accountable - a department lead or enablement owner |
| `intervention_type` | interned: briefing / scenario workshop / champion session / comms / one-to-one / licence reassignment |
| `guidance_resource` | optional link to the Microsoft resource used (#555) |
| `started_utc`, `due_utc`, `completed_utc`, `status` | |
| `intended_outcome` | **`nvarchar`** free text - what success looks like |
| `notes` | `nvarchar` |

Then surface, per intervention: the cohort's trajectory across subsequent periods (#544), and eventually the difference-in-differences result (#550).

## Why this shape

It deliberately keeps the *execution* outside the product - we are not building a task manager. The record exists so the **measurement** has something to attach to: without a date and a membership list, "did it work?" is unanswerable.

## Acceptance criteria

- An intervention can be created from the enablement-plan drill-through in one step, freezing the cohort as part of the same action.
- Overdue and unstarted interventions are visible.
- The intervention appears alongside the cohort's outcome figures, so the claim and the evidence sit together.
- Free-text columns are `nvarchar`.
- Additive, idempotent, guarded migration with a matching `.manual.sql`.

## Notes

- Depends on #548.
- From the September 2026 Copilot Adoption review, plan item 3.2.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

The intervention itself is recorded against the frozen cohort.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

Issue #544, opened 2026-09-13
Copilot Adoption: cohort progression - retained, reactivated, lapsed and reassigned


sambetts:
## Summary

The question that actually proves an enablement programme worked cannot be asked today:

> *"Of the 300 seats that were **Never used** last quarter, how many are active now? How many were reclaimed? How many are still idle?"*

This is the single most valuable output of the whole adoption area for an executive audience, and it is a self-join once #542 exists.

## Proposed transitions

Between two closed periods, classify every seat that existed in the earlier one:

| Transition | Meaning |
| --- | --- |
| **Retained** | Active in both periods |
| **Reactivated** | Inactive (never-used or dormant) then, active now - *the enablement win* |
| **Lapsed** | Active then, inactive now - *the churn signal, currently invisible* |
| **Reclaimed / reassigned** | Held a seat then, no seat now - *proof the reclaim happened* |
| **Newly assigned** | No seat then, holds one now |
| **Still at risk** | Inactive in both periods - *the hard core* |

Report as counts, as a flow (the transitions between bands), and with a drill-through to each population per wiki principle 2.9 - filtered on the same key the aggregate grouped by.

## Why this is differentiated

Microsoft's own Copilot Dashboard **cannot answer the reclaim question at all**. From its FAQ:

> *"employees who **only** have a Copilot license are **removed from all analyses in the dashboard, regardless of the time period, after their Copilot license is removed**."*

Microsoft erases the evidence at the moment the action is taken. A tool that retains it can show a reclaim programme worked; Microsoft's cannot.

## Acceptance criteria

- Transitions are computed between any two published periods and always sum to the earlier period's population.
- Each transition drills through to exactly its members.
- "Reclaimed" is distinguishable from "user left the organisation" where the data allows, and the ambiguity is stated where it does not.
- Cohort figures appear in the workbook so a before/after can be presented without the portal.
- Transitions respect the comparability gate (#543).

## Notes

- Depends on #542, and on #277 for reliable seat-removal detection (a failed or partial import must never be read as mass licence removal).
- From the September 2026 Copilot Adoption review, plan item 2.4. All three reviewing models named cohort progression as the real answer to the progress-tracking principle.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

Cohort progression shipped: retained, reactivated, lapsed and reclaimed.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

PR #586, opened 2026-09-16
Split Copilot Adoption into executive and analyst views

sambetts:
## Summary

Addresses #552.

Splits the Copilot Adoption Overview into two first-class views:

- **Executive view**: a three-act briefing (`Where we stand`, `Where it is working and failing`, `What we are doing about it`) focused on licence position, reclaim confidence, department habit gaps, the funnel, and the enablement plan.
- **Analyst view**: retains the existing workbench diagnostics and SQL popovers, and repeats the executive headline/detail bridge so executive figures can be checked without reconstructing them.

## Implementation notes

- Replaced the old `overview` tab state with `executive` and `analyst` tabs.
- Added an `ExecutiveTab` composed from existing primitives (`KpiGrid`, `GaugeRing`, `CategoryBarChart`, `AdoptionFunnel`, `ActionPlan`) rather than adding new API data.
- Added a department league table ranked by habit rate. It shows the non-colour encodings needed for board-pack use: licence count, never-used seats, recommended candidates, and the reassignment signal in text.
- Kept InfoTips in both views, but kept `</> SQL` controls out of the Executive view. SQL remains on the Analyst view cards for query-backed figures.
- Kept the existing first-run/zero-licence empty state as the Executive view empty state instead of showing an empty diagnostic grid.
- Added drill-through from executive actions to the Licensed users tab using the same action code used by the aggregate.

## Additive seams left for parallel work

No placeholder values were invented for branches that are not available here:

- #543 period deltas: `buildExecutiveKpis` now centralises the executive KPI subset, so delta display can be added to those KPI definitions without restructuring the view.
- #553 idle spend: the reclaim confidence card is isolated and can accept currency once the API provides it.
- #549 owners: the enablement-plan card already owns the executive action surface and can add owner metadata beside each action.
- #544 cohort outcomes: the same enablement-plan card is the insertion point for last-period outcome summaries.

## Tests

- `npm run test`
- `npx tsc --noEmit`

## Risks / reviewer hotspots

- The Executive department league derives its reassignment signal from existing department adoption and opportunity summaries. Review the wording to ensure it does not overclaim purchase-vs-reassignment economics before #553 lands.
- The Analyst view deliberately retains the workbench volume. The new Executive view is the C-level entry point; the Analyst view is for verification and diagnosis.

## Wiki updates required

Proposed replacement for wiki section 5:

```markdown
## 5. Page structure: Executive view and Analyst view

The Copilot Adoption page is split into two views because the tool has two readers. The budget owner needs a board-pack answer first; the admin or enablement lead still needs the full workbench and the SQL behind it.

### 5.1 Executive view

The Executive view is a one-screen briefing in three acts:

1. **Where we stand** — assigned Copilot seats, adoption rate, habit rate, reclaimable licences split by confidence, and the reassignment opportunity from recommended unlicensed candidates.
2. **Where it is working and failing** — a department league table ranked by habit rate, showing licence count, never-used seats, recommended candidates and the reassignment signal in text, plus the adoption funnel.
3. **What we are doing about it** — the enablement plan. Each row names the next step and drills through to the exact licensed-user population counted by that row.

The Executive view keeps inline definitions and InfoTips. It deliberately does **not** show `</> SQL` buttons, because a SQL button next to a board-pack KPI reads like unfinished software to a C-level reader. The working is still available in the Analyst view.

### 5.2 Analyst view

The Analyst view is the verification and diagnostic workbench. It keeps the existing charts and tables: engagement mix, frequency buckets, intensity scatter, score-profile radar, concentration cohorts, app/resource breakdowns, weekly trends, country and combined licensed/unlicensed breakdowns, and the SQL popovers for query-backed figures.

Any number promoted to the Executive view must also be reachable from the Analyst view or a detail tab. A headline must not exist only as a board-pack assertion.

### 5.3 Detail tabs

The Licensed users, Cowork, Unlicensed usage, Agents and Licence opportunities tabs remain the operational drill-through surfaces. Aggregates that name a population must filter by the same key used to build the aggregate. For example, an enablement-plan row groups by recommended-action code, so its drill-through filters by recommended-action code.

### 5.4 Empty states

The Executive view owns the first empty-state explanation. A genuine zero-licence result should explain the likely causes in order: licence data not imported yet, no SKU classified as a Copilot licence, or the tenant genuinely having no Copilot licences. It should point the reader to the Health page, licence-type settings, or Licence opportunities as appropriate. It must not render a wall of empty diagnostic cards.
```


sambetts:
Fixed both review findings in this branch.

- Added a separate `habitByDepartment` summary list, ranked server-side by lowest habit rate, so the executive Department league table is no longer limited to departments preselected by adoption rate.
- Updated the executive card copy/InfoTip to state both the seat-threshold omission and the top-segment truncation.
- Trimmed `OpportunityByDepartment` grouping keys so candidate counts use the same department label domain as adoption segments.

Validation:
- Enterprise MSBuild Release restore/build for `src\AnalyticsEngine\Tests.UnitTests`: passed.
- `vstest.console.exe /TestCaseFilter:"FullyQualifiedName~CopilotAdoption"`: 233 passed.
- `npm ci --loglevel verbose --timing --no-audit --no-fund`: passed.
- `npm run test`: 212 passed.
- `npx tsc --noEmit`: passed.
- `git --no-pager diff --check`: passed.

sambetts:
## Adoption funnel visibility fix

Fixed the additional executive-view funnel visibility finding. The funnel component itself is pre-existing on `dev`; #586 did not introduce the low-contrast label bug, but it made the component prominent by promoting it into the Executive/board-pack view, so the fix belongs in this PR.

What changed:
- Added a WCAG relative-luminance helper that composites each stage fill over the white card at the gradient edge alpha (`0.8`) before choosing the in-shape label colour.
- The count and `% of licensed` caption now use whichever of white or `tokens.colorNeutralForeground1` has better contrast for that stage colour, so future palette changes keep labels legible without a hardcoded per-index override.
- Removed the blanket caption dimming for dark neutral labels; opacity `0.85` is kept only when white is selected on a dark fill.
- Left the outside-label path and SVG hit areas/tooltips unchanged; the existing tooltip tests still pass.

Contrast before/after, measured at the conservative 0.8-alpha edge over white:

| Stage | Raw fill | Painted edge | Before: white text | After label | After contrast |
|---|---:|---:|---:|---|---:|
| Licensed | `#8ec3ea` | `#a5cfee` | 1.65:1 | dark neutral | 9.43:1 |
| Ever used Copilot | `#5aa6dd` | `#7bb8e4` | 2.14:1 | dark neutral | 7.26:1 |
| Active this period | `#2f86cc` | `#599ed6` | 2.88:1 | dark neutral | 5.39:1 |
| Habitual users | `#1466ad` | `#4385bd` | 3.94:1 | white | 3.94:1 |
| Champions | `#0a4a80` | `#3b6e99` | 5.42:1 | white | 5.42:1 |

Also checked the right-hand drop column: Fluent `colorPaletteRedForeground1` (`#bc2f32`) on `colorNeutralBackground1` (`#ffffff`) is 5.85:1, so the existing 12px red "N lost here" text meets the 4.5:1 target and still carries meaning in words as well as colour.

Validation:
- Added the contrast unit test first against the hardcoded-white helper and confirmed it failed (`expected '#ffffff' to be '#242424'`).
- `npx vitest run src\components\copilotAdoption\AdoptionFunnel.test.tsx`: 4 passed after the fix.
- `npm run test`: 213 passed.
- `npx tsc --noEmit`: passed.
- `git --no-pager diff --check`: passed.

sambetts:
## Adoption funnel contrast correction

Corrected the funnel-label contrast fix to measure against the raw stage fill instead of the earlier 0.8-alpha gradient edge. The label is centred around the fully painted part of the band, so raw fill / approximately 0.98 alpha is the right measurement point; this keeps the two dark stages white and changes only the three pale/mid bands that need dark text.

Corrected contrast table, using raw stage fill and dark `#111111` as the neutral foreground reference:

| # | Stage | Fill | vs white | vs dark | Correct pick |
|---:|---|---:|---:|---:|---|
| 0 | Licensed | `#8ec3ea` | 1.88:1 | 10.02:1 | DARK |
| 1 | Ever used Copilot | `#5aa6dd` | 2.65:1 | 7.13:1 | DARK |
| 2 | Active this period | `#2f86cc` | 3.89:1 | 4.86:1 | DARK |
| 3 | Habitual users | `#1466ad` | 5.94:1 | 3.18:1 | WHITE |
| 4 | Champions | `#0a4a80` | 9.11:1 | 2.07:1 | WHITE |

The helper remains luminance-driven rather than index-driven, still returns `tokens.colorNeutralForeground1` for the dark option in the component, and still removes the blanket 0.85 opacity from dark captions. Outside-shape labels and `pointerEvents: 'none'` on text were left unchanged.

Validation:
- Temporarily restored the hardcoded-white helper and confirmed the corrected test failed (`expected '#ffffff' to be '#111111'`).
- `npx vitest run src\components\copilotAdoption\AdoptionFunnel.test.tsx`: 4 passed after restoring the luminance implementation.
- `npm run test`: 213 passed.
- `npx tsc --noEmit`: passed.
- `git --no-pager diff --check`: passed.

PR #583, opened 2026-09-16
Add Copilot Adoption licence spend reporting

sambetts:
## Scope

Addresses #554
Addresses #553

Implements the Copilot Adoption licence-spend story as one backend/API/portal/workbook change:

- reports purchased, assigned, unassigned and assigned-idle Copilot seats per classified SKU;
- keeps purchased/unassigned as `Unknown` when Graph `subscribedSkus` is unavailable rather than rendering zero;
- adds optional per-SKU cost inputs (`monthly` or `annual`, currency, effective date) and produces idle-spend figures only when those inputs are supplied;
- separates spend exposure, reassignable assigned-idle seats and reducible-at-renewal unassigned seats;
- keeps all currency figures as separate `AzureCostByCurrency` rows and never sums currencies together.

### Import / data model

- Reuses the existing user metadata import path that already calls Graph `subscribedSkus`; no new Graph fetch/import loop was added.
- Adds `202609170920001_CopilotSubscribedSkuCapacity`, storing `subscribedSkus.prepaidUnits` on `dbo.license_types` as nullable capacity columns plus `subscribed_sku_refreshed_utc`.
- Clears those nullable capacity fields when tenant-level `subscribedSkus` is unavailable, so the report explicitly becomes unknown instead of showing stale or zero purchased-seat data.
- Extends the licence classification contract with `purchasedUnits`, `unassignedUnits`, `assignedIdleUsers` and `purchasedUnitsRefreshedUtc`.

### Adoption analysis / API

- Adds optional `CopilotSeatCostInput` values to `CopilotAdoptionOptions` and includes them in the analysis cache key.
- Builds `IdleLicenceSpendSummary` with:
  - `spendExposure` = assigned eligible idle spend + unassigned spend;
  - `reassignable` = eligible assigned-idle spend only;
  - `reducibleAtRenewal` = unassigned spend only;
  - tier breakdowns keyed by the existing reclaim confidence tiers.
- Requires cost, currency, period and effective date before producing a monetary figure.

### Portal / workbook

- Adds per-SKU seat-cost entry controls to the Copilot Adoption page.
- Adds a purchased-seat KPI and an idle-spend summary card.
- Expands the SKU methodology table to show assigned, purchased, unassigned and assigned-idle seats.
- Adds the new figures to the workbook headline/method sheets.

### Installer prerequisite coverage

- Updates installer Test Configuration coverage text/checks for the user metadata import to include `Organization.Read.All`, because `subscribedSkus/prepaidUnits` requires it.

[...]

Issue #558, opened 2026-09-13
Cowork: ingest the first-party usage report, and model eligibility as a spending policy rather than a licence


sambetts:
## Summary

Cowork support today is a counted flag: `CoworkUsers`, `CoworkAdoptionPct`, `CoworkInteractions`, `CoworkDetected`, a per-user boolean, a CSV column, a trend series and a filter. Research into the current product documentation shows three things that make that model wrong in ways worth fixing.

## 1. Our unit of measurement disagrees with Microsoft's

Microsoft counts **tasks**, defined as:

> *"A **task** represents a **single conversation thread** within Cowork, including both user-initiated conversations and scheduled task executions. **If a user continues an existing task by asking additional questions, it counts as the same task.**"*

We count audit **interactions**. These differ by a large factor, in the direction that makes our figures look inflated. A customer comparing our number with the admin centre will conclude we are wrong.

Cowork is also an **agentic system** - Microsoft is explicit that it is *"different from an agent"* - that runs *"minutes to hours (autonomous execution)"* of multi-step work. So placing `CoworkInteractions` beside Copilot Chat interaction counts, as the tool does now, invites a comparison that is meaningless in both directions.

## 2. There is a documented first-party Cowork usage report - we should ingest it

**Admin centre -> Copilot -> Cowork.** Per-user, CSV-exportable, documented, with data back to **1 April 2026**.

- Tenant metrics: active Cowork users, total tasks, average tasks per active user, **retained Cowork users**, grace-period countdown.
- Per-user: `Total tasks`, **`Scheduled tasks`**, **`User-initiated tasks`**, `Active days`, `Last activity date`.

This is stable and documented, unlike our audit heuristics (see 4 below). It should be the basis for Cowork adoption figures.

## 3. Eligibility is granted by a SPENDING POLICY, not a licence

There is **no Cowork SKU**. Cowork requires the Microsoft 365 Copilot USL *plus* usage-based billing, is **off by default**, and access is controlled by spending policy:

> *"**A spending policy is an access control, not only a budget.** Any user in the scope of a spending policy that selects Cowork can use Cowork, **regardless of how small the credit limit is. A policy with a limit of one credit still grants access.**"*

And the control we currently read is deprecated:

> *"The **agent-based access control from the Frontier and Preview versions is deprecated**... The **Cowork** agent entry is still visible under **Agents > All Agents**, but **any configuration on it has no effect** on who can use Cowork."*

[...]

PR #587, opened 2026-09-16
Add Copilot adoption accountability rollup

sambetts:
## Summary

Addresses #556 by adding a configurable Copilot Adoption accountability roll-up that defaults to direct manager and can be pointed at department, country, office, or company through `CopilotAdoptionOptions.AccountabilityDimension`.

## Implementation

- Adds `AccountabilityRollupRow` to the summary API with seats, active users, habitual users, adoption rate, reclaim confidence tiers, and counts for each recommended action.
- Defaults the dimension to direct manager and explicitly groups users with no manager as `(no manager)`.
- Reuses the existing `MinSeatsPerSegment` threshold so small leader groups are suppressed exactly like department/country segments.
- Sorts accountable groups by absolute opportunity (`reclaim`, `win back`, `coach`, `broaden`, `deepen`, `review`) before percentages so larger fixable spans surface first.
- Adds the roll-up to the portal overview and to the Excel workbook export.

## Governance position (#538)

Issue #538 would have added a per-user access role for managers, but it has been closed as not planned. This PR therefore does not add a new named per-user disclosure surface for leaders. The leader/accountability view is aggregate-only by default, applies the existing `MinSeatsPerSegment` suppression to prevent arithmetic disclosure for tiny spans, and leaves existing licensed-user drill-through behaviour as the ceiling for named detail.

## Performance and scale

- No database schema change or index is added, so there is no migration or index benchmark requirement.
- Incremental roll-up work is one pass over the already-scored licensed-user result and one in-memory grouping by a single allow-listed dimension. At the current `MaxLicensedUsersScored` cap that is O(50k) rows, small compared with the existing SQL scoring inputs. A future uncapped/exact 200k-seat aggregate should move the equivalent grouping into a dedicated SQL aggregate before raising that cap.

## Tests

- `MSBuild.exe src\AnalyticsEngine\Common\Entities\Entities.csproj -t:Restore,Build -p:Configuration=Release -p:Platform=anycpu -m:2 -v:m`
- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -t:Restore -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -m:2 -v:m`
- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -m:2 -v:m`
[...]

PR #588, opened 2026-09-17
Source Cowork adoption from the first-party usage report

sambetts:
Addresses #558

## Summary

Cowork adoption was a counted flag derived entirely from audit heuristics. Three things about that model were wrong against Microsoft's current documentation, and this branch fixes all three.

1. **Our unit of measurement disagreed with Microsoft's.** Microsoft counts *tasks* — one conversation thread, where continuing an existing task stays the same task. We counted audit *interactions*. These differ by a large factor in the direction that makes our figures look inflated, so a customer comparing our number with the admin centre concludes we are wrong.
2. **Eligibility is granted by a spending policy, not a licence.** There is no Cowork SKU. The agent-based access control we were reading (`CoworkAgentIdsSql`) is deprecated — the agent row is still visible and still returns rows, so nothing looked broken, but configuration on it no longer governs access. `CoworkAdoptionPct` therefore divided by all licensed users, which is the wrong denominator.
3. **A documented first-party per-user usage report exists** and is a stable basis, unlike our undocumented `app_host = 'cowork'` / agent-id-prefix heuristics.

### Import
- `CoworkUsageUserDetailLoader` ingests the per-user Cowork usage report: `Total tasks`, `Scheduled tasks`, `User-initiated tasks`, `Active days`, `Last activity date`.
- Registered in `ProductionGraphImportSectionFactory` alongside the existing Copilot usage-report import; `CopilotReportRequest` extended for the new report.

### Model and scoring
- Report-sourced task figures are preferred for adoption; the audit-derived interaction figure is retained and **kept visibly distinct** for reconciliation, never presented as comparable with Microsoft's task count.
- Eligibility is modelled as *USL AND in scope of a Cowork-selecting spending policy*. Nothing reads the deprecated agent entry to infer eligibility.
- `CoworkAdoptionPct` is **suppressed when eligibility cannot be established** rather than divided by the wrong population.
- New metrics: **automation ratio** (scheduled / total tasks — a delegation/maturity signal rather than a usage one) and **7-day retention**, available per user and per department.

### Schema
- Migration `202609170940001_CoworkUsageReportTables` — **purely additive** (new tables only; no column removal, no type change, no index on an existing table). Per the repository rule, purely additive schema carries no before/after benchmark because there is no prior query to measure.
- Idempotent and guarded, with a matching `202609170940001_CoworkUsageReportTables.manual.sql` whose pre-stamp guard checks **schema only, never data state**, and which falls through to the stamp on the already-applied path.
- All customer-originating text columns are `nvarchar`.
- Data-retention SQL (`Clean Data By User StoredProc.sql`, `Clean Old Data Data.sql`) updated so the new tables are covered by existing cleanup rather than growing unbounded.

## Tests

- `CopilotAdoptionCoworkTests` extended: report-sourced figures preferred with audit retained separately; adoption percentage suppressed on unknown eligibility rather than mis-divided; eligibility never inferred from the deprecated agent entry; interaction counts never labelled as task counts.
- `UrlFullUrlMigrationPipelineTests.LatestId` updated to `202609170940001_CoworkUsageReportTables`.

**Local validation:** Release build clean. `Cowork` + `Migration` filtered run: **127 passed, 2 failed**. Both failures are environmental, not from this change — `HealthServiceTests.Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` and `CopilotEventManagerAccessedResourcesTests` fail in `Guid.Parse` because `App.Release.config` still holds the unsubstituted CI placeholder `__TenantGUID__` on a developer machine. Verified by running the same test on an unrelated branch (`sambetts/users-last-updated-utc`), where it fails identically.

## Risks and reviewer hotspots

- **Eligibility inference is the hotspot.** When spending-policy scope is unavailable the figure is suppressed rather than estimated. Reviewers should confirm suppression is what surfaces, since a wrong denominator here is worse than no number.
- The audit heuristics are retained deliberately for reconciliation — the agent-id prefix match and the `ag.name LIKE '%Cowork%'` fallback still exist so divergence from the report becomes visible instead of silent.
- Cowork audit records carry no provider details, so no model breakdown is available from audit data; none is claimed.

## Deferred

- **Period-fact carry-through (#542).** `cowork_interactions` and the new task counts should accrue into the period-fact table so Cowork history exists. #542 is being built in parallel and is not in this branch; a clean seam is left for it.
- **Credits per task.** Cowork is billed in Copilot Credits and is the first Copilot workload with a real attributable unit cost. Seat-cost reporting is landing separately in #583 (#553/#554); wiring credits-per-task into that existing `AzureCostByCurrency` model — including its "never add across currencies" rule — is better done once, after that lands, than duplicated here.

## Wiki updates required

The wiki is a shared checkout, so it is not edited in this PR. Required changes to `Copilot Adoption Tool.md`:

- State that Cowork adoption is sourced from the first-party usage report, with the audit-derived figure retained only for reconciliation.
- Document that Microsoft counts **tasks** (one conversation thread, continued questions stay the same task) and that our interaction counts are **not** comparable — and are deliberately not presented as such.
- Replace any description of Cowork eligibility as licence-based: there is no Cowork SKU; access comes from a spending policy that selects Cowork, and *any* policy in scope grants access regardless of how small the credit limit is.
- Note that the Cowork entry under **Agents > All Agents** is deprecated and its configuration has no effect on access, so it must not be used to infer eligibility.
- Document the new automation-ratio and 7-day-retention metrics and when the adoption percentage is suppressed.


sambetts:
Fixed both review findings in commit `e3e978f7`.

- **MAJOR 1 — fixed by dropping the workbook column:** removed the per-department **Cowork adoption %** column from the Cowork rollout-order workbook section. The workbook now keeps the counts admins can act on plus the UI-aligned automation/coordination/fluency fields, and no longer presents a percentage whose denominator is departmental licensed seats rather than spending-policy eligibility.
- **MAJOR 2 — fixed:** removed the `ag.name LIKE '%Cowork%'` fallback from `CoworkAgentIdsSql`, so only product-controlled first-party `agent_id` prefixes feed the observed Cowork usage path. Restored/strengthened tests to prove a tenant-owned agent named `Contoso Cowork Helper` is not returned and cannot promote users into evidence tiers or remove them from candidates.

Validation:
- VS18 Enterprise `MSBuild` restore/build for `src\AnalyticsEngine\Tests.UnitTests` Release x86 — passed.
- `vstest.console.exe ... /TestCaseFilter:"FullyQualifiedName~Cowork|FullyQualifiedName~CopilotAdoption" /Platform:x86` — 236 tests passed.
- `npm run test` — 28 files / 208 tests passed.
- `npx tsc --noEmit` — passed.

Note: `npm ci` populated packages but hung before creating executable links in this worktree, matching the #587 worktree behaviour. I repaired the local validation install with `npm rebuild` and a direct reinstall/extract of `@fluentui/react-icons@2.0.330`; no package manifests were changed.

sambetts:
## Release-assembly note: migration snapshot ordering across the four parallel migrations

Flagging a cross-PR hazard found while integrating this release. **No action is needed on this PR in isolation** — it is correct as it stands — but it must be checked once the migration-bearing PRs are merged.

### The situation

This release currently carries four new migrations, on top of `202609161200001_IndexTeamsExplorerQueries` which is already on `dev`:

| Run order (migration id) | PR | Changes the EF model? | How its `.manual.sql` stamps |
|---|---|---|---|
| `202609170900001_CopilotAdoptionPeriodFacts` | #579 | **No** — new tables are not `DbSet`s | copies predecessor's `Model` |
| `202609170910001_GraphCopilotUsageApiV2` | #580 | **Yes** — adds `report_version` | embeds its own freshly scaffolded blob |
| `202609170920001_CopilotSubscribedSkuCapacity` | #583 | **No** | copies predecessor's `Model` |
| `202609170940001_CoworkUsageReportTables` | #588 | **No** | copies predecessor's `Model` |

Verified by decoding each `.resx` `Target` and comparing against the `dev` baseline: only #580's differs (baseline 46,122 bytes).

### The hazard

Each branch's snapshot is correct *relative to its own branch*, because each was cut from a `dev` that did not contain the other three. But the migrations are applied in **id order**, and #580 — the only one that changes the model — sits **second**. The two raw-SQL migrations that run after it (`...0920001`, `...0940001`) copy a predecessor row whose `Model` predates `report_version`.

The consequence is not in the schema — every table and column is created correctly — it is in what `__MigrationHistory` *claims* the model is. If the highest-id migration ends up stamped with a pre-`report_version` model while the built entity model contains it, EF sees the schema as behind the build. Concretely that surfaces as:

- `HealthServiceTests.Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` failing in CI, and
- at runtime, `Migrations/Configuration.cs` having `AutomaticMigrationsEnabled = true`, so EF would auto-generate an on-the-fly migration rather than treating the schema as current.

This only bites when more than one of these four merges, which is why no individual PR shows it.

### What to check at merge time

After the migration-bearing PRs are merged to `dev`:

1. Confirm the **highest-id** migration's snapshot matches the final merged entity model. If #580 has merged, that snapshot must include `report_version`.
2. Treat `Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` as the gate — it is the check that actually proves it, and it runs in CI. (It cannot be run on a developer machine: `App.Release.config` ships the unsubstituted placeholder `__TenantGUID__`, so that whole test class fails in `Guid.Parse` locally. ~161 such failures appear on every branch, including untouched ones.)
3. If it fails, re-scaffold the highest-id migration's `.resx` against the merged model and update its `.manual.sql` stamp to match, rather than relaxing the test.

### Why the manual scripts themselves are still safe

Each `.manual.sql` names a predecessor that is already on `dev`, so the severity-16 prerequisite guard cannot spuriously fire regardless of the order these four land in. The stamp guards were separately verified to be **schema-only** (`OBJECT_ID` / `COL_LENGTH` / `sys.indexes`), with the already-applied path falling through to the stamp rather than returning early — so re-running any of them is a no-op, and none of them can refuse to stamp because of data state.

sambetts:
Resolved the `origin/dev` merge conflicts for this branch and pushed the result.

Integration notes:
- Kept both sides of the migration registrations in `Entities.csproj` for 202609170900001, 202609170910001, and 202609170940001.
- Kept `LatestId` at `202609170940001_CoworkUsageReportTables`, with the dev-side period-facts / Graph v2 notes and this branch's Cowork note in order.
- Confirmed the 202609170900001, 202609170910001, and 202609170940001 migrations each have `.cs`, `.Designer.cs`, and `.resx` registrations in `Entities.csproj`.
- Updated `202609170940001_CoworkUsageReportTables.manual.sql` to point at predecessor `202609170920001_CopilotSubscribedSkuCapacity`, per the release migration chain.
- Fixed the raw-SQL 202609170940001 migration snapshot to reuse the post-Graph v2 model snapshot, preventing EF from auto-generating a duplicate `report_version` column during migration-pipeline tests.

Validation:
- Conflict-marker scan across `*.cs`, `*.tsx`, `*.ts`, `*.csproj`, `*.sql`: clean.
- MSBuild restore/build, Release/anycpu/x86: passed.
- Filtered VSTest (`CopilotAdoption|Period|UrlFullUrlMigration|Cowork`): 288 passed / 0 failed.
- Portal Vitest: 254 passed / 0 failed.
- Portal TypeScript (`tsc --noEmit`): passed.

sambetts:
Follow-up snapshot check after `202609170910001_GraphCopilotUsageApiV2` landed on `dev`.

- Compared `202609170940001_CoworkUsageReportTables.resx` with `202609170910001_GraphCopilotUsageApiV2.resx`: it already matched from the earlier integration fix, so no further source change was needed.
- The intended predecessor is `202609170920001_CopilotSubscribedSkuCapacity`; because that raw-SQL migration now reuses Graph v2's snapshot, this branch's matching `Target` keeps the chain consistent end to end.
- Manual SQL stamps by copying `Model` from predecessor `202609170920001_CopilotSubscribedSkuCapacity`, so no embedded blob regeneration was needed.
- Built Release test project and ran the migration-pipeline test against private LocalDB catalog `UnitTestingSPOInsights_merge583588`: `UrlFullUrlMigration` 3 passed / 0 failed.

Issue #547, opened 2026-09-13
Copilot Adoption: customer-defined adoption targets (internal goals, not industry benchmarks)


sambetts:
## Summary

Nothing on the page has anything to compare against. Every figure is absolute, so a reader cannot tell whether 46% adoption is a success or a failure for *this* organisation.

## Explicitly: internal targets, not external benchmarks

This issue is deliberately scoped to **customer-defined goals**. External/industry benchmarks are out of scope and should stay that way:

- a self-hosted product has no lawful, representative cross-tenant dataset;
- Microsoft *does* ship internal and external benchmarks in the Copilot Dashboard, but they are **in-product only and gated at 50+ licences**, and Microsoft publishes **no numeric targets** anyone could hard-code;
- all three models reviewing this area independently recommended against external benchmarks.

If an external anchor is ever wanted, the only defensible Microsoft-sourced numbers are its **Power user** definition (>=15 actions/week in >=9 of 12 weeks) and **Habitual user** definition (1-14/week, same consistency rule) - and those must be attributed to Microsoft and kept separate from our own bands (see the vocabulary collision in #531).

## Proposed

A target is: a metric, a scope (whole tenant / department / cohort), a target value, an owner, a baseline period and a target date.

- Show progress against target next to the figure, with the **baseline frozen** so the goalposts cannot move retrospectively.
- Targets belong on the factual metrics - recurring use, activation rate (#546), reduction in idle seats - rather than on a composite score, so hitting a target cannot be achieved by retuning a threshold.
- If `options_hash` changes between baseline and now, restate the baseline from stored facts (#542) or flag the target as not comparable.

## Acceptance criteria

- A target records its baseline period and value at creation, and that baseline is immutable.
- Progress is shown against the target with the owner named.
- Retuning `CopilotAdoptionOptions` cannot silently move a target's measured progress.
- Targets appear in the workbook and the digest.

## Notes

- Depends on #542 for a stable baseline.
- From the September 2026 Copilot Adoption review, plan item 2.7.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

Customer-defined targets with frozen baselines. Baselines are immutable at the database level, so target progress cannot be quietly improved by retuning the scoring options after the fact.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

PR #590, opened 2026-09-17
Add Copilot adoption deltas and targets

sambetts:
> **Stacked PR:** this is stacked on PR #579 (`sambetts/copilot-adoption-period-facts`) and must merge after it. Base branch is intentionally `sambetts/copilot-adoption-period-facts`, not `dev`.

Addresses #543
Addresses #547

## Summary
- Adds closed-period movement for every Copilot Adoption headline KPI, with selectable comparison (`previousPeriod` or `samePeriodLastQuarter`).
- Movement is calculated only from published closed period facts; the current partial period is never used as either side of the delta.
- Refuses deltas when the published periods have different `options_hash` values, and includes scored-seat denominator movement next to rate deltas.
- Adds customer-defined internal targets with frozen baseline period/value/options hashes, owner, scope, target value and target date.
- Surfaces target progress in the portal and writes both movement and targets into the Excel workbook.

## Implementation notes
- Reuses #542's `copilot_adoption_period_run` / `copilot_adoption_user_period` read path and comparability gate rather than adding a parallel history source.
- Touched #542 files deliberately (`CopilotAdoptionPeriodModels`, `CopilotAdoptionService`, `CopilotAdoptionSql`, workbook summary paths) to extend the period-fact model with movement/targets.
- The target baseline is immutable: the migration adds `dbo.TR_copilot_adoption_targets_freeze_baseline`, and the API has no update path for baseline fields.
- Targets store both the fact-shaping `options_hash` and a full scoring-options hash. Retuning `CopilotAdoptionOptions` marks target progress not comparable instead of silently moving the baseline.
- Portal overlap to watch: `CopilotAdoptionPage.tsx` and workbook files are also active in concurrent work (#586 / workbook changes). The changes here are limited to comparison controls, KPI hints and target display.

## Schema / migration
- Migration: `202609171000001_CopilotAdoptionTargets`.
- Classification: **purely additive** (new empty target table, lookup index and immutability trigger). No performance-motivated before/after benchmark is required.
- Manual DBA script included: `202609171000001_CopilotAdoptionTargets.manual.sql`.
- Manual script prerequisite: `202609170900001_CopilotAdoptionPeriodFacts`; it stamps by copying that predecessor snapshot and verifies schema only before stamping.

## Tests
- `MSBuild src\AnalyticsEngine\Tests.UnitTests -t:Restore -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `MSBuild src\AnalyticsEngine\Tests.UnitTests -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:m -m:2`
- `vstest src\AnalyticsEngine\Tests.UnitTests\bin\Release\Tests.UnitTests.dll /TestCaseFilter:"FullyQualifiedName~CopilotAdoption|FullyQualifiedName~Period"` — 264/264 passed
- `npm run test` — passed
- `npx tsc --noEmit` — passed

## Wiki updates required
- Update the Copilot Adoption Tool page to add the third product question: “is our enablement programme working?”
- Document that movement uses stored, closed period facts only and refuses non-comparable `options_hash` pairs.
- Document internal customer targets, baseline immutability, owner/scope/target-date fields and the absence of external/industry benchmarks.
- Add the manual upgrade note for `202609171000001_CopilotAdoptionTargets.manual.sql` and its prerequisite order after #579's period-facts migration.


sambetts:
## Fixed during integration verification: the migration was invisible to EF

Flagging a defect found and fixed while verifying this branch, because it would not have shown up in the validation this PR originally reported.

### What was wrong

`202609171000001_CopilotAdoptionTargets.Designer.cs` declared the migration id as:

```csharp
string IMigrationMetadata.Id
{
    get { return "202609170900001_CopilotAdoptionTargets"; }   // predecessor's timestamp
}
```

That is the **predecessor's timestamp** (`202609170900001`, from `CopilotAdoptionPeriodFacts`) paired with this migration's name — the signature of a copied designer file whose class name was updated but whose id was not.

### Why it matters

The id in the designer, not the file name, is what EF matches on. With the wrong timestamp:

- `UrlFullUrlMigrationPipelineTests` upgraded only as far as `202609170900001_CopilotAdoptionPeriodFacts`, and
- on a real customer upgrade through `DatabaseUpgrader.CheckDbUpgraded`, **the targets tables would never have been created** — silently, with no error, because EF simply did not see a newer migration to apply.

The feature would have appeared to ship and then not worked, with the failure surfacing only when someone tried to create a target.

### Why the original validation missed it

The reported run used `/TestCaseFilter:"FullyQualifiedName~CopilotAdoption|FullyQualifiedName~Period"`, which does not match `UrlFullUrlMigrationPipelineTests` — the one test that actually walks a database up the migration chain and asserts it lands on `LatestId`. `LatestId` itself was correctly updated to `202609171000001_CopilotAdoptionTargets`, so nothing in the touched-file set looked wrong.

This is the case the repository guidance calls out as the easiest to miss: *"the class name mentions neither 'migration id' nor the area you're working in, so a filtered local test run usually won't cover it."*

### The fix

One line — the declared id now matches the file name and `LatestId`:

```csharp
get { return "202609171000001_CopilotAdoptionTargets"; }
```

**Verified:** Release build clean; `CopilotAdoption|Period|UrlFullUrlMigration` now passes **267/267** against a freshly dropped `UnitTestingSPOInsights` database. Before the fix the same filter failed 2 tests with `Expected:<202609171000001_CopilotAdoptionTargets>. Actual:<202609170900001_CopilotAdoptionPeriodFacts>. DB should now be at the latest migration.`

I also audited every other migration designer on all the release branches for the same copy-paste defect — `202609170900001`, `202609170910001`, `202609170920001`, `202609170940001`, `202609171020001` and `202609171020002` all declare ids matching their file names. This was the only one.

### Note for future migrations on this release

Run the migration pipeline test explicitly — a `CopilotAdoption`-shaped filter will not cover it:

```powershell
& $vstest ...\Tests.UnitTests.dll /TestCaseFilter:"FullyQualifiedName~UrlFullUrlMigration"
```

and drop `UnitTestingSPOInsights` first, because every worktree shares one LocalDB instance and concurrent runs contaminate it.

PR #592, opened 2026-09-17
Base the agent cost report on verified billing dimensions, and capture the Cowork service tag

sambetts:
## Why

The Copilot Costs page was built partly against a billing schema that was assumed rather than observed. This PR checks each panel against live Microsoft billing data, removes the dimensions nothing can populate, and adds the one dimension that actually separates Copilot Cowork spend.

Evidence was gathered against a real tenant and a real Azure subscription: the Cost Management `query` API, the Power Platform licensing API under both a delegated Global Administrator and the importer's own application-only principal, and the network payloads served to the Microsoft 365 and Power Platform admin centres. No tenant identifiers, resource names or figures from that work appear in this diff.

## Azure cost import

**Capture the tag a query was grouped by.** Microsoft bills Copilot Cowork, Work IQ API and Copilot Studio through a *single* `Pay As You Go Copilot Credit` meter under the `Microsoft Copilot Studio` service, and documents no per-experience line item. The only discriminator is a tag on the usage record — and note it is injected by the billing pipeline, **not** an ARM tag, so it does not appear on the resource in the portal.

- New `azure_cost_daily.tag_key` / `.tag_value`, populated from the `TagKey` / `TagValue` columns a tag grouping returns.
- New `tag` Azure breakdown dimension.
- Opt in with `AzureCostGroupBy=ResourceId;TagKey:serviceName`. Nothing changes until an operator sets it.

The trade-off is real and was measured, not assumed: Cost Management permits only two group-by clauses, so a tag grouping consumes one. With it enabled, `meter_name` is **not** returned and goes NULL. `AzureCostQueryParser` resolves tags by the fixed `TagKey`/`TagValue` column names only — aliasing the tag's own name would collide with the identically-named dimension, because the column index is case-insensitive.

The four fields are part of the Azure row hash. With a tag grouping the same resource and meter returns once per tag value, so omitting them would collapse those rows and sum them, destroying the split the grouping was configured to get.

**Summary quantity is now conditional.** `AzureCostByCurrency.Quantity` is populated only when a single meter/tag combination contributes. Quantities of different meters are different units — credits, GB, hours, operations — and the naive sum rendered as `5,342,761 metered units` beside the money on a real subscription, which reads as a credit count and is not one. For an unfiltered subscription it is now null, and the per-tag or per-meter breakdown is where a quantity is scoped to one unit.

## Copilot Studio credit report

**Removed four dimensions with no obtainable source.** A live capture of the per-agent credit response returned a `metadata` object of exactly `ResourceName`, `NonBillableQuantity` and `Users`. The LLM model, tool invoked, knowledge sources and channel appear in **no Microsoft REST reference and in no observed payload**, so every pivot, grid column, filter and CSV column backed by them could only ever answer "Not reported" — which an admin reads as "this tenant did none of that".

Removed from the reporting surface *and* the schema: `AgentCostDimensions`, `AgentCostDetailRow`, `AgentCostFilterOptions`, `AgentCostQuery`, the controller parameters, the TypeScript types, the page, the CSV export, and the `copilot_studio_credit_daily` columns.

**Added a credit-side availability guard.** `AgentCostAvailability.CreditDimensionsWithData` mirrors the existing `AzureDimensionsWithData`: a pivot is offered only when there is data behind it, and returns by itself if Microsoft starts sending it. Harness counts as present only when something was genuinely classified — `NotAssessed` and `Unknown` both mean "could not tell", and a pivot whose only bucket is that is a dead end.

**Kept fields now accept both documented and observed names.** The REST reference documents `metadata` as *"Feature, ProductName and nonBillableConsumed"*; the parser was written against `FeatureName`, `ResourceName`, `NonBillableQuantity`. Those are different names, not different casings, and matching was case-insensitive only — so reading one set would have yielded null for ever, taking the agent name, the billing feature and (because harness is classified from the feature) the harness with it. The live capture confirmed the *observed* spellings are correct and the reference is stale, but neither source is authoritative — the same reference also documents a flat envelope live tenants do not return. Both are accepted.

## Fixes

**ISO-dated cost responses were silently discarded on non-US-culture hosts.** Newtonsoft parses an ISO-8601 string into a `Date` token; `ToString()` then rendered it in the host culture (`15/02/2026` on en-GB) and the subsequent `InvariantCulture` parse rejected it, so the row had no usable date and was dropped. On such a host an entire ISO-dated response became "no spend at this scope". Now read as a date, normalising a `Local` kind to UTC first so the day does not shift west of UTC.

**Reworded the credit-import failure message.** It led with "assign a Power Platform role", which sends an admin who has already done so round in circles. Verified on a live tenant: the licensing routes refuse an application-only token whose service principal *already holds* Power Platform Reader at tenant scope, and the per-agent route is refused even to a signed-in Global Administrator. The message now names both possibilities.

## Migrations

Two, forming a strict chain. Both are metadata-only and need no maintenance window on any edition.

| Migration | Class | Effect |
|---|---|---|
| `202609171120001_AzureCostTagColumns` | **Additive** | Two NULLable `nvarchar` columns on `azure_cost_daily`. Instant at any table size. |
| `202609171125117_DropUnreportedCopilotStudioCreditColumns` | **Removal** | Drops four unpopulated `copilot_studio_credit_daily` columns. No index, constraint or default references them, so no rebuild. |

They are the last two links of the chain, in that order, immediately after `202609171030001_CopilotAdoptionDigest`.

The tag migration was renumbered from `202609161425295` when `dev` was merged in. `dev` had landed eight migrations in the meantime, which left the original id sorting in the middle of them — breaking `EveryManualScript_NamesItsImmediatePredecessor`, which requires each manual script to name the migration directly before it in executable SQL. Both model snapshots were rebased onto `dev`'s model at the same time: the chain head's snapshot is the one EF compares the live entity model against, and the pre-merge one would have had EF auto-migrate an `ADD report_version` for a column `dev` had already added. Both `.manual.sql` model blobs were regenerated from the rebased snapshots and verified byte-identical to their `.resx`.

Neither is performance-motivated, so no before/after benchmark is owed — stated explicitly rather than left implicit.

Both ship a guarded, idempotent `.manual.sql` that stamps `__MigrationHistory`. The stamp guards check **schema only**, never row state. **CI uploads `**/*.zip` only, so both `.manual.sql` files must be attached to the release by hand.**

**Two things a reviewer should weigh:**

1. `copilot_studio_credit_daily` **has shipped** — it went out in stable build 1830 and again in 1832, so it exists in customer databases. This is a `DROP COLUMN` against live schema, not a new table. It is still the right call because the values are unreachable from the source API and are NULL everywhere observed, and nothing about money, credits or attribution is lost. A confirming query is included in the script's header comment.
2. The dropped fields were part of `dimension_hash`. The shortened tuple means an existing slice **inserts once** under its new key rather than updating, then settles from the following run. That is one duplicate per slice per day, once — not ongoing drift — and zero for any deployment whose credit import has never succeeded, which is every one observed.

## Residual uncertainty

The agent in the captured response had `consumed = 0`, and the admin centre requested only `includeFields=users`. So the evidence that those four fields never arrive is strong but not absolute — it cannot fully separate *"these do not exist"* from *"these appear only with real consumption"*. Re-adding them later is a UI plus additive-migration change, not a rewrite. Generating real Copilot Studio consumption against prepaid capacity would settle it at no Azure cost.

## Testing

- `.NET`: after the `dev` merge, 2542 passed / 2558 across the whole suite, with the migration chain replayed from a dropped LocalDB. The 12 failures are all environmental on a clean worktree — placeholder tenant credentials (`AADSTS90002`), an unreachable cognitive endpoint, and a missing local `InstallerTestConfig.json` — and none touch this change.
- `UrlFullUrlMigrationPipelineTests` pass, including the `LatestId` assertion now pointing at the new head.
- Portal: `tsc --noEmit` clean, 191 vitest tests pass.
- Both `.manual.sql` scripts validated against a scratch database: apply from the prior state, stamp with the correct model blob, and are a no-op on re-run. EF then reports **0 pending migrations**, proving each stamped snapshot matches the live entity model. Re-validated after the merge, from a database downgraded to `dev`'s head.

New regression tests cover: the tag grouping being sent as a `TagKey` clause; `TagKey`/`TagValue` parsing without colliding with `ServiceName`; untagged rows surviving; rows differing only by tag staying separate; both metadata spellings; the culture-dependent ISO date; the summary quantity staying null across multiple meters; and the tag pivot resolving more than one service.

The Greek charset sample moved from `knowledge_sources` to `agent_name` (`nvarchar(255)`) in the demo generator and its three assertions, so the Unicode round-trip still crosses a real encoding boundary rather than being quietly lost with the dropped column.

Addresses the agent-cost reporting accuracy work.


sambetts:
## Blocked on a migration-snapshot conflict with the Copilot Adoption release — not merged

I attempted to merge this into `dev` as part of preparing the `dev` → `main` release and **stopped**, because it breaks the migration chain when combined with the nine migrations that landed today. The branch is untouched; nothing was pushed to it.

### What happens

Merging this branch and running the migration pipeline test fails both of its from-legacy-state upgrades:

```
Microsoft.Data.SqlClient.SqlException: Column names in each table must be unique.
Column name 'report_version' in table 'dbo.copilot_usage_user_activity_log' is specified more than once.
```

`UrlFullUrlMigrationPipelineTests` walks a database up the entire migration chain, so this is the upgrade path a customer takes.

**Control:** current `dev` alone passes that same test **3/3**. The failure appears only once this branch is merged, so it is an interaction between the two sets of migrations, not a pre-existing defect on either side.

### Why

`202609170910001_GraphCopilotUsageApiV2` (merged in #580) does `AddColumn("dbo.copilot_usage_user_activity_log", "report_version", …)`. With `AutomaticMigrationsEnabled = true`, EF reconciles each migration's model snapshot as it walks the chain, and this branch's two snapshots were scaffolded against a model that does not match the merged one — so EF auto-generates an `ADD report_version` that then collides with the explicit one.

I tried two orderings and both fail, so it is not simply a sequencing problem:

1. **As authored** — `202609161425295_AzureCostTagColumns` sorts *before* `…0900001_CopilotAdoptionPeriodFacts`, yet its snapshot already contains `report_version`, which does not exist in the database at that point in the chain.
2. **Renumbered** to `202609171120000`, placing it after `…1030001_CopilotAdoptionDigest` so the snapshot's contents match its position. Same failure.

### What I verified along the way (all fine — no action needed)

- All 11 migrations register correctly in `Entities.csproj` (`Compile` ×2 + `EmbeddedResource` ×1 each).
- Snapshot sizes are internally coherent: Digest 664,009 → AzureCostTagColumns 664,252 (+ the two tag columns) → DropUnreported 663,128 (− the four dropped columns).
- Neither branch's snapshots contain the new Copilot Adoption tables, which is **correct** — those are raw-SQL and not exposed as `DbSet`s, so they are deliberately absent from the EDMX.
- The `AgentCostsPage.tsx` conflict is straightforward: this branch removes the filters for the columns it drops, and that side should win.

### What this branch needs

**Re-scaffold both migrations' `.resx` snapshots against merged `dev`**, using `Add-Migration -Force` (or the reflection-based scaffolder) once `dev` contains the release. Then:

- Confirm the **last** migration's snapshot matches the current merged entity model — that is what EF compares against, and what `Config_PendingMigrations_MeanTheSchemaIsBehindTheBuild` checks.
- Re-point the `.manual.sql` prerequisite guards and stamp sources at the **immediate** predecessor in the merged order. As merged the chain ends `…1020002_CopilotAdoptionInterventions` → `…1030001_CopilotAdoptionDigest`, so these two follow it.
- Consider whether `AzureCostTagColumns` should be renumbered to sort after the adoption migrations; its content assumes a later model than its current id implies.
- Re-run `/TestCaseFilter:"FullyQualifiedName~UrlFullUrlMigration"` — it must be **3/3**.

### Release status

The Copilot Adoption release (20 issues, 9 migrations) is verified green and is proceeding to `main` without this PR. This is not a judgement on the change itself — the billing-dimension work looks sound and its own CI passed — only that its migration snapshots predate today's chain and must be regenerated before it can merge safely.

Happy to do the re-scaffold if you would like, but it belongs to whoever owns this branch, since it needs a rebuild against the merged model.

Issue #557, opened 2026-09-13
Copilot Adoption: scheduled aggregate digest (no per-user data, no attachments)


sambetts:
## Summary

Everything in the adoption area is pull-based: someone has to remember to open the portal. Executives do not, which means the tool's findings reach them only when someone else does the work.

## Proposed

A scheduled digest - monthly by default - containing:

- headline figures **with their period-over-period deltas** (#543);
- progress against any defined targets (#547);
- cohort outcomes from the last period (#544, #550);
- the top two or three actions by size, with owners where interventions exist (#549);
- a link into the portal for detail.

## Hard constraints

- **Aggregate only. No named individuals, ever.** A per-user emailed digest is an uncontrolled disclosure channel that also persists in mailboxes indefinitely, outside any access control the product has. The review was unanimous on this.
- **A link, not an attachment.** The recipient authenticates and gets whatever their role permits (#538) - so the digest cannot become a way around the role split.
- **Ships after #538 and #543.** A digest of untrustworthy numbers just distributes the problem faster, and a digest without deltas has nothing to say.
- Respect `MinSeatsPerSegment` in any segment mentioned.

## Acceptance criteria

- The digest contains no UPN, display name or any per-user row, in body or attachment.
- Recipients are configurable, and the digest states the period and the options/threshold set used.
- Sending is idempotent - a retry does not double-send.
- Failure to generate is visible in the Health area rather than silent.
- A deployment that has not configured it sends nothing.

## Notes

- Depends on #538 and #543.
- From the September 2026 Copilot Adoption review, plan item 4.6.

sambetts:
Shipped in **Stable build 1833** (merged to `main` via #596).

Scheduled aggregate digest shipped - no per-user data and no attachment, just a link into the tool, so it cannot route named data past access control.

https://github.com/pnp/Microsoft365-Analytics-Insights/releases/tag/1833

Issue #538, opened 2026-09-13
Copilot Adoption: per-user adoption data has no role separation - default the surface to aggregate-only


sambetts:
## Summary

This is a **data-protection design gap**, not a remotely exploitable vulnerability: the endpoints require authentication. But once authenticated, **any** portal user can read and export per-individual productivity data about every colleague, and nothing records that they did.

`Web/Controllers/CopilotAdoptionAPIController.cs:45` is decorated:

```csharp
[Authorize]
```

There is **no `[Authorize(Roles = ...)]` and no `IsInRole` anywhere in the `Web` project** - a repository-wide search returns zero matches. Every other controller is the same, so this is a product-wide architectural gap that happens to matter most here, because this is the most sensitive dataset the product produces.

## What is exposed

Through `licensed-users`, `licensed-users/export`, `opportunities`, `opportunities/export` and `export/workbook`:

- name, UPN, mail, department, job title, country, office, **manager**, account status;
- interactions, active days, days since last use, engagement score and band;
- an **adverse per-user action verb** - *"Reclaim or onboard - no Copilot activity in the last 365 days. Confirm the licence is still needed before renewal."*

Volume limits are generous by design: `MaxCsvRows = 100000` (`CopilotAdoptionAPIController.cs:59`) and `MaxUserRows = 20000` in the workbook (`CopilotAdoptionWorkbook.cs:27`).

`MinSeatsPerSegment` (`CopilotAdoptionOptions.cs:279`, default 5) k-anonymises the **segment charts**. The per-user drill-through bypasses it entirely.

## Why this needs addressing

Three things compound:

1. The flagship reclaim number currently produces **named false positives** (#532) - it can assert that a person on parental leave is not using their licence.
2. That named list is **exportable by any authenticated user, with no audit trail**.
3. The stated audience for this tool explicitly includes **Microsoft account teams** - a third party.

In several EU jurisdictions, surfacing individual productivity metrics to management - and more so to a third party - can require works-council agreement and a DPIA. Disclosure to a vendor raises purpose-limitation questions. The product currently ships no access control, no data minimisation and no audit, so **the customer carries all of that risk with no tooling to manage it**.

The wiki specification already states the intended position as a non-goal:

> *"**Not** a per-user surveillance tool."*

The intent is right. **Nothing in the code enforces it.**

## Suggested direction

[...]

PR #591, opened 2026-09-17
Add Copilot adoption intervention cohorts

sambetts:
## Stacked PR

**This PR is stacked on #579 / `sambetts/copilot-adoption-period-facts` and must merge after it.** The base branch is intentionally `sambetts/copilot-adoption-period-facts`, not `dev`, because the implementation depends on the #542 period-fact publish/read path and options-hash comparability gate.

Addresses #548
Addresses #549
Addresses #550
Addresses #551

## What landed

- Adds frozen Copilot Adoption cohorts and immutable cohort membership with T0 band, score, active days, department and optional hold-out-control flag.
- Adds intervention records attached to a frozen cohort: owner, interned intervention type/status, optional guidance resource code, dates, intended outcome, notes, and intended reinvestment intent.
- Adds API endpoints to create a cohort or intervention from an enablement-plan action, list/export members, close cohorts, list interventions, and read an intervention outcome.
- Adds difference-in-differences outcome measurement against either a random hold-out or matched untreated users (baseline band, 10-point score bucket, department; tenure explicitly deferred until #277).
- Adds guarded small-group refusal using the existing `MinSeatsPerSegment` philosophy instead of emitting a noisy effect.
- Adds leading-indicator movement for Teams messages/meetings, email sent/read, and SharePoint/OneDrive file activity against the matched control. No ROI, money, hours-saved or causation claim is emitted.
- Wires the portal enablement plan with a one-click “Start intervention” action and fixes existing strict TypeScript implicit-any handler errors so `tsc --noEmit` is green.

## Schema / migrations

- `202609171020001_CopilotAdoptionCohorts` — **purely additive**: new cohort/member tables, constraints and table-local keys only.
- `202609171020002_CopilotAdoptionInterventions` — **purely additive**: new intervention table and constraints only.
- Matching `.manual.sql` scripts are included for both migrations; their pre-stamp guards check schema only, never row data.
- EF model snapshots are unchanged; both migrations reuse the #542 snapshot and stamp by copying their predecessor.
- No installer config schema change. No performance-motivated schema change; no benchmark required.

## Deferred / follow-up

- Rich portal CRUD for editing intervention owner/dates/notes after creation. The backing API/data model supports the fields now; this PR adds the one-step creation path from the enablement plan and outcome endpoints.
- Guidance-resource picker UI pending the #555 guidance catalogue surface being available in this stacked branch.
- Tenure matching remains explicitly out of scope until #277 provides assignment history.

## Tests

[...]

PR #593, opened 2026-09-17
Add scheduled Copilot Adoption digest

sambetts:
## Stack / merge order

**This PR is stacked on #590 (`sambetts/copilot-adoption-deltas-targets`), which is stacked on #579 (`sambetts/copilot-adoption-period-facts`). Merge order must be: #579 -> #590 -> this PR.**

Addresses #557.

## Summary

- Adds an opt-in scheduled Copilot Adoption digest from the Office365ActivityImporter cycle.
- Sends aggregate-only HTML via the existing Graph app identity path: no workbook/CSV attachment and no per-user rows.
- Reuses the existing Copilot Adoption period publishing/read path, #543 period deltas, #547 targets, action-plan summaries, and existing `MinSeatsPerSegment`-suppressed segment summaries.
- Adds digest send-state storage so retries do not double-send a period/recipient set and failed generation/send attempts surface in Health.

## Privacy / governance

- No default recipients and no default sender: deployments send nothing until `CopilotAdoptionDigestRecipients` and `CopilotAdoptionDigestSenderUserId` are explicitly configured as App Service app settings.
- The digest renderer takes only `CopilotAdoptionSummary`; it never renders `LicensedUsers`, UPNs, mail/display-name fields, or per-user exports.
- #538 was closed **not planned**, so its stated gate cannot be satisfied. This PR supplies the required protection for email by making the digest aggregate-only, link-only, and non-attachment-based; recipients must authenticate in the portal for detail and get only what their role permits.

## Schema / configuration

- Adds migration `202609171030001_CopilotAdoptionDigest` and matching `.manual.sql`.
- Migration classification: **purely additive** operational table (`dbo.copilot_adoption_digest_run`) plus indexes; no performance benchmark gate applies.
- Manual script is idempotent, copies the predecessor snapshot, and the pre-stamp guard checks schema only.
- No installer saved-config schema change; `CONFIG_VERSION` is not bumped. Digest settings are unmanaged App Service app settings and are preserved by existing installer merge behaviour.

## Tests

- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -t:Restore -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `MSBuild.exe src\AnalyticsEngine\Tests.UnitTests -p:Configuration=Release -p:Platform=anycpu -p:ProcessorArchitecture=x86 -v:q -m:2`
- `vstest.console.exe src\AnalyticsEngine\Tests.UnitTests\bin\Release\Tests.UnitTests.dll /TestCaseFilter:"FullyQualifiedName~CopilotAdoption|FullyQualifiedName~Period|FullyQualifiedName~Digest"` — 275/275 passed
- `cd src\AnalyticsEngine\Web\Scripts\portal; npm run test -- --run; npx tsc --noEmit`

## Wiki updates required

- Document the new optional App Service settings: `CopilotAdoptionDigestRecipients`, `CopilotAdoptionDigestSenderUserId`, and `CopilotAdoptionDigestIntervalHours` (default 720 hours / monthly).
- State that the digest is aggregate-only, sends no attachment, respects `MinSeatsPerSegment`, and links to the portal for role-controlled detail.
- Note that the runtime app registration/mailbox must have the Graph `Mail.Send` capability appropriate for the configured sender.


sambetts:
## Fixed: `test_dotnet (Release)` failure — 4 tests, one root cause

CI run [35226467528](https://github.com/pnp/Microsoft365-Analytics-Insights/actions/runs/35226467528) failed with 4 of 2,472 tests failing. All four had the same cause, introduced by this PR.

| Failed test | |
|---|---|
| `AppConfigDefaultsTests.NoAppSettingsAtAll_ReproducesLegacyStableBehaviour` | direct |
| `GraphImportTests.AllTeamsLoadTest` | collateral |
| `GraphImportTests.CallQueueProcessorTest` | collateral |
| `GraphUsageReportImportTests.SharePointSitesUsageLoaderTest` | collateral |

All four threw `System.FormatException: Unrecognized Guid format` from `AppConfig..ctor()`.

### Root cause

This PR added `TenantGUID` to `AppConfigDefaultsTests._trackedKeys` and set it in `[TestInitialize]`:

```csharp
ConfigurationManager.AppSettings.Set(TenantGUID, "00000000-0000-0000-0000-000000000000");
```

That breaks two things.

**1. The direct failure.** `TenantGUID` is the *only* setting the `AppConfig` constructor parses with a hard `Guid.Parse` rather than a `TryParse` (`AppConfig.cs:26`) — every other knob tolerates a missing or junk value by design. The pre-existing upgrade-parity test `NoAppSettingsAtAll_ReproducesLegacyStableBehaviour` iterates `_trackedKeys` and blanks every one of them to simulate a deployment that has none of the new settings. Adding `TenantGUID` to that list meant the test now blanked it too, and the constructor threw instead of falling back.

**2. The three collateral failures.** `ConfigurationManager.AppSettings` is **process-global**, and `tests.yml` runs vstest with `runInParallel: true`. While `AppConfigDefaultsTests` held `TenantGUID` blanked or overwritten, any test running concurrently that constructs an `AppConfig` threw as well. That is exactly what happened to the three Graph tests — they have nothing to do with the digest.

The setting was only ever added to work around the *local* developer-machine condition where `App.Release.config` still holds the unsubstituted placeholder `__TenantGUID__`. CI substitutes the real value from `secrets.APP_TENANTGUID` before the build, so CI never needed it — the workaround imported a local-only problem into the shared test run.

### The fix

Reverted both additions. The diff for this file is now purely the two new digest test methods and their key constants; nothing mutates `TenantGUID`.

### Verification, in both directions

Because the failure is invisible on a developer machine (where the whole class fails on the placeholder anyway), I reproduced CI's conditions locally by substituting a valid GUID into the untracked `bin\Release\Tests.UnitTests.dll.config`:

- **Fixed code:** `AppConfigDefaults` filter → **39/39 pass**, including `NoAppSettingsAtAll` and both new digest tests.
- **Defect reintroduced** (adding `"TenantGUID"` back into `_trackedKeys`, nothing else): `NoAppSettingsAtAll` fails with the identical `System.FormatException: Unrecognized Guid format` — confirming causation rather than correlation.
- **Full targeted suite** on the fixed code, against a freshly dropped `UnitTestingSPOInsights`: `AppConfigDefaults|Digest|CopilotAdoption|Period|UrlFullUrlMigration` → **315/315 pass**.

### Note for anyone touching `AppConfigDefaultsTests`

`_trackedKeys` drives a test that deliberately blanks every key in it. Only add a key there if `AppConfig` degrades gracefully when it is missing. `TenantGUID` does not — it is the single hard-parse in the constructor. And because the settings collection is process-global while CI runs tests in parallel, mutating a key that other tests depend on will surface as unrelated failures elsewhere in the run.

sambetts:
Rebased onto current `origin/dev` and force-pushed.

Resolved integration conflicts from the portal split and the moving migration chain:
- Kept the new `ExecutiveTab`/`AnalystTab` split; did not restore `OverviewTab`.
- Preserved the interventions wiring from #591 in the stacked tree: `onCreateIntervention` and `interventionMessage` flow into `ExecutiveTab`'s enablement-plan `ActionPlan`.
- Preserved both TenantGUID fixes; there is no `ConfigurationManager.AppSettings.Set("TenantGUID", ...)`, and `TenantGUID` is still absent from `AppConfigDefaultsTests._trackedKeys`.
- Unioned the stacked migration registrations for subscribed SKU capacity, Cowork usage-report tables, targets, cohorts/interventions, and digest; verified `Entities.csproj` parses as XML.
- Updated the migration test chain so digest remains latest: `202609171030001_CopilotAdoptionDigest`.
- Corrected `202609171000001_CopilotAdoptionTargets`, `202609171020001_CopilotAdoptionCohorts`, `202609171020002_CopilotAdoptionInterventions`, and `202609171030001_CopilotAdoptionDigest` `.resx` `Target` snapshots to match `202609170910001_GraphCopilotUsageApiV2`.
- Pointed digest manual SQL/test predecessor at `202609171020002_CopilotAdoptionInterventions`; cohorts points at targets; targets points at Graph v2.
- Confirmed no conflict markers remain.

Validation on the rebased tip:
- MSBuild Restore + Build (`Release`): passed.
- Filtered VSTest (`CopilotAdoption|Period|UrlFullUrlMigration|Digest`) using `UnitTestingSPOInsights_rebase`: 321 passed / 321 total.
- Portal `vitest run --reporter=basic`: 40 files passed, 257 tests passed.
- Portal `tsc --noEmit`: passed.

PR #598, opened 2026-09-18
Make the Copilot Adoption per-user evidence readable

sambetts:
## Scope

Three reported readability defects on the Copilot Adoption tabs. All are portal-only — no API, no schema, no migration, no config change. Six files under `Web/Scripts/portal/src`.

## 1. `InfoTip` popover (`components/shared/InfoTip.tsx`)

**Defect.** `surface.maxWidth` was `380px`. The monospace Calculation block is `white-space: pre-wrap`, so the engagement formula re-wrapped mid-expression:

```
depth      = min(1, interactions / activeDays /
depthTarget) x confidence
```

A formula that re-wraps reads as a different formula, which defeats rendering it verbatim.

**Change.**
- `maxWidth: 'min(560px, calc(100vw - 32px))'`. The widest line the engagement tip can emit is ~66 chars ≈ 507px including both paddings, so it fits with margin; the `calc` keeps it inside a narrow viewport.
- New `formulaBlock` class (`white-space: pre` + `overflow-x: auto`) applied via `mergeClasses` **only when `content.formula` contains a newline**. Single-line formulas are frequently prose-like (`"12 certain + 8 probable seats …"`) and must keep wrapping — turning those into a horizontal scroller would be a regression. `mergeClasses` rather than string concatenation because the two classes both set `white-space`, so Griffel has to resolve the conflict deterministically.
- `how` now renders in a `<div>` carrying `fontSizeBase200`/`lineHeightBase200` instead of Fluent's `<Text>`. `TextSlots.root` is `'span' | 'h1'..'h6' | 'p' | 'pre' | 'strong' | 'b' | 'em' | 'i'` — no `div` — so a `<ul>` inside it was invalid nesting. ~40 existing call sites pass a plain string and are visually unchanged.

## 2. Engagement score tooltip (`components/copilotAdoption/LicensedUsersPanel.tsx`)

**Defect.** `confidence` appeared in the displayed calculation but was named nowhere in the prose — the paragraph said only "Depth is additionally scaled down below 3 active days". The term was therefore undefined at its point of use. The narrative form exists only in the page's Engagement score accordion, which is the wrong place: by the time a reader has the question they are looking at the column, not the methodology tab.

**Change.** Reworked into three bulleted components (`How often — frequency`, `How much — depth`, `How widely — breadth`, each carrying its weight and target) plus two short paragraphs, one of which defines `confidence` explicitly.

Correctness notes for review:
- The scaling wording is **target-agnostic**: "below that it counts in proportion, so half as many days counts for half". That holds for any `DepthMinActiveDays`, unlike "1 day counts a third", which is only true at the default of 3.
- Formula now uses `frequency`/`depth`/`breadth` (matching `CopilotAdoptionScoring.Score`) instead of the previous mix of `freq` and undefined `depthTarget`/`breadthTarget` placeholders, and substitutes the configured target values.
- The `/ weightSum` divisor is omitted when `Math.abs(weightSum - 1) < 1e-9`. With default weights it rendered as `/ 1 x 100`, which reads as noise; when weights are customised the divisor is still shown, so the displayed formula stays faithful to `Score`.

## 3. Per-row expanders on Cowork readiness + Licence opportunities

**Defect.** Both tables overflow horizontally, and both ended with a `Justification` column that was *simultaneously* clamped to two lines (`RationaleCell`, `-webkit-line-clamp: 2`) and off the right-hand edge. The per-user sentence — the artefact an admin pastes into a licence request or rollout plan — was the one thing on the row that could not be read, and the supporting Microsoft 365 usage figures had the same problem.

**Change.** Justification column removed from both tables; a chevron expander added instead.

New shared primitives in `components/copilotAdoption/adoptionShared.tsx`:

| Export | Purpose |
|---|---|
| `useRowExpansion` | `ReadonlySet<number>` keyed **by user id**, so a re-sort cannot leave a different person's detail open. Exposes `collapseAll`. |
| `ExpandableUserCell` | Returns the identity `<td>` with the toggle |
| `DetailRow` | Full-width `<tr>` / `colSpan` `<td>` |
| `DetailSections` / `DetailSection` / `DetailStats` / `DetailStat` | Layout for the figures |
| `DetailRationale` | The justification, unclamped |

Design decisions worth a reviewer's attention:

- **The chevron lives inside the sticky-left User cell, not in a leading column of its own.** A separate leading column would have to be sticky too, which means hard-coding its rendered width into the `left:` offset of the User column — or it scrolls out of reach on exactly the wide tables the expander exists to rescue.
- **`detailInner` is `position: sticky; left: 0` with a bounded `width: min(1040px, calc(100vw - 96px))`.** The detail cell spans every column, so it is as wide as the whole overflowing table; left as a normal block the panel would scroll away with the columns and reproduce the original defect. The bounded width is what gives `sticky` something to pin.
- **`detailColSpan` tracks the conditional column.** CoworkPanel renders "All Copilot Credits" only when `credits?.perUserCreditsAvailable`, hence `? 9 : 8`. Declared after `const credits = summary.coworkCreditPosition` to avoid a TDZ error.
- **`useEffect(() => collapseAll(), [filters, windowDays, page, collapseAll])`** in both panels. Paging or re-filtering replaces the rows under an open detail, so without this the expander would describe whoever lands on that line next. `collapseAll` is `useCallback`-stable.
- Existing hover `Tooltip`s on the score bars are **retained** — the expander adds a route to the numbers, it does not remove the old one.
- Nullable fields keep their "not attributable ≠ zero" semantics: `coworkReportTotalTasks`, `coworkReportActiveDays`, `coworkAutomationRatioPct` and `coworkCreditsPerTask` each render their `DetailStat` only when non-null, and credits use `formatCredits`, never `formatCount`.

**Removals.** `RationaleCell` and its `rationale` clamp style are deleted — nothing referenced them once both columns were gone. `OpportunitiesPanel`'s now-unused `InfoTip` import and `thWithInfo` style are removed with it.

**Stale copy fixed.** Two InfoTip strings in `OpportunitiesPanel` referred to "The Justification column on each row" and "Hover any row for its four component scores"; both now point at the expander.

## Tests

`adoptionShared.test.tsx`: the two `RationaleCell` tests are replaced by five expander tests. The load-bearing one asserts the justification is rendered **in full and without a `title` attribute** — i.e. it pins the exact regression this change fixes, rather than the clamp-and-hover behaviour it replaces. Others cover collapsed-by-default, collapse-on-second-click, and that expanding one row does not expand another.

## Verification

- `npx tsc --noEmit` — clean
- `npx vitest run` — 40 files, 264 tests, all passing
- Multi-model critique loop run before push

## Reviewer hotspots

1. `detailColSpan` 8 vs 9 against the real `<th>` count in `CoworkPanel`.
2. Whether `detailInner`'s sticky/bounded-width trick behaves on your screen width — it is the one piece that is easier to judge running than reading.
3. Removing the Justification column outright, rather than keeping a truncated copy in the row. The full text remains in the CSV/Excel export, unchanged.


sambetts:
## Review round applied (commit `293c55b9`)

Ran a multi-model critique loop over the first commit. Three reviewers; one returned clean, two found the same "Qualified by" issue plus three others. All four findings were in code this PR adds, all verified against the C# source before changing anything, and all are now fixed.

| # | Finding | Verified against | Fix |
|---|---|---|---|
| 1 | `Reported tasks` rendered the scheduled / user-initiated split as `?? 0`. Those two columns are independently nullable and blank means "not reported", not "none" — the same distinction the per-user credit column protects. | `CoworkReadinessRow` in `types/copilotAdoption.ts`; `CopilotAdoptionSql.cs` | `taskSplitLabel` names only the halves actually reported, otherwise "split not reported". Guard test added. |
| 2 | `Last Cowork use` read `lastCoworkInteractionUtc`, which is audit-only — a report-only user (which the scorer treats as **observed**) showed a dash while still being badged observed. The report date was also hidden whenever `coworkReportActiveDays` was null, because it was nested under that stat. | `CopilotAdoptionSql.cs:876-892` derives `LastCoworkInteractionUtc` from the audit CTE and `CoworkReportLastActivityDate` from `report.LastActivityDate` separately | Relabelled `Last audit interaction`; the report date is now an independent stat with its own source label. |
| 3 | `Qualified by` re-derived the route client-side against the raw configured `opportunityProvenDemandMinActiveDays`. The scorer floors it with `Math.Max(1, …)`, so a non-positive setting would label **every** row "Proven demand", including users with no Copilot use. | `CopilotAdoptionScoring.cs:1055-1061` | The server already computes and serialises `qualificationTierLabel` (`CopilotAdoptionUserModels.cs:464-469`, camelCase `JsonProperty` confirmed present — this project has no camelCase resolver, so that mattered). Added both fields to `LicenceOpportunityRow` and render the server's label. One source of truth instead of two that agree only at the default. |
| 4 | `detailInner` was bounded by `calc(100vw - 96px)`, which ignores the left navigation and page padding. On a 1200px viewport that resolves to the 1040px cap inside a scrollport of roughly 870px, so part of the panel stayed clipped — `sticky` pins the left edge, it does not shrink an over-wide element. | Reported by review; arithmetic checked against the page chrome | `tableWrap` now declares `container-type: inline-size` and the panel is bounded by `100cqw`, i.e. the scrollport itself. Griffel emits it fine and the panels render in the suite. |

**Not changed, deliberately.** One reviewer asked for the tooltip to also describe the scorer's degenerate-config branches (`weightSum <= 0 ⇒ score 0`, `Ratio` special-casing a non-positive target). Those displayed expressions are unchanged from `dev` and are only reachable by misconfiguring an option to zero or less, so fixing them here would be an unrelated pre-existing change; flagging rather than folding it in.

**Re-verified after the fixes:** `tsc --noEmit` clean, full portal suite 40 files / 266 tests passing (two more than before — the new guards).

